PR #295, opened 2013-05-22
Decouple filtered methods detection and test instances initialization from suite executions.

calavera:
The interface in Minitest 5 makes extending Runnable harder. The `run` method in the `Runnable` class has three responsibilities:

1. Detect methods to run.
2. Create test instances.
2. Run tests.

Moving the detection logic and the test instances initialization to separated methods makes it more easy to extend.


zenspider:
This case is no longer valid. Closing.

PR #339, opened 2013-08-09
Initial implementation of prove_it flag for tests.  If success is not pr...

bradleyjames:
Exception based handling of unit tests have always bothered me as by design they assume success.  This commit changes that in minitest so that the developer is notified that they haven't proven anything if they haven't asserted a condition.  My preference would be to enable this by default but that may prove to be too aggressive for some at this point in time.

This is a proposal for inclusion into minitest, it is not yet complete.  I have not added unit tests or updated the unit tests to use this flag.  If you feel it should be included in the core I'll follow up with more commits.  If not I'll look into building a plugin.

### DIFF
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -14,6 +14,11 @@ module Minitest
     PASSTHROUGH_EXCEPTIONS = [NoMemoryError, SignalException, # :nodoc:
                               Interrupt, SystemExit]
 
+    class << self
+      attr_reader :prove_it
+    end
+    @prove_it = false # for backwards compatibility must be false
+
     ##
     # Call this at the top of your tests when you absolutely
     # positively need to have ordered tests. In doing so, you're
@@ -52,6 +57,15 @@ module Minitest
         undef_method :test_order if method_defined? :test_order
         define_method :test_order do :parallel end
       end
+    end
+    
+    ##
+    # Call this at the top of your tests when you want to force all
+    # tests to prove success (via at least one assertion) rather 
+    # than rely on the absence of failure.
+    
+    def self.prove_it!
+      @prove_it = true
     end
 
     ##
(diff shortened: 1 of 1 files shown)


zenspider:
Man... I still really want this.

teoljungberg:
@zenspider If it's of any interest, I would gladly extract this into a plugin

zenspider:
I _still_ really want this... *grrrr*

zenspider:
I've extracted this out to `minitest/proveit.rb`:

```ruby
class Minitest::Test
  class << self
    attr_accessor :prove_it
  end

  ##
  # Call this at the top of your tests when you want to force all
  # tests to prove success (via at least one assertion) rather
  # than rely on the absence of failure.

  def self.prove_it!
    @prove_it = true unless defined?(@prove_it)
  end

  TEARDOWN_METHODS << "assert_proven?"

  def assert_proven? # :nodoc:
    flunk "Absence of failure is not success. Prove it!" if
      self.class.prove_it && self.failures.empty? && self.assertions == 0
  end
end

# to be moved to minitest/hell?
class Minitest::Test
  def self.run reporter, options = {}
    prove_it!
    super
  end
end
```

I've currently got it wired to forcibly turn on for everyone, but I'm looking to move that to `minitest/hell.rb`. I have run this through my stuff several times now and had to fix a number of things (esp generated tests and minitest/mock tests). `minitest-gcstats` intentionally has tests w/o assertions so it goes off there unless I disable it (opt-out!). So I think this is a happy addition to the system.

I want an easier way to globally turn on `minitest/hell` just to make things even more fun.

zenspider:
GOD I LOVE THIS SO MUCH

zenspider:
OK. I'm still fussing... but consider this merged. It's gonna happen, damnit.

teoljungberg:
It looks like this hasn't been pushed and merged yet, I would love to run this through my projects and see what damage it can do.

zenspider:
I'm about to release it as a gem. It'll have hooks in minitest/hell to try to activate it if you have it installed.

PR #368, opened 2013-11-25
add `expect` method that returns an expectation proxy

tenderlove:
Add an `expect` method so that we don't have to rely on thread locals for the test instance.  If we deprecate the object monkey patches (as RSpec has done) than we can completely eliminate the thread locals.

Also, [as noted](https://github.com/seattlerb/minitest/blob/dfa67e4ce0f94ceab2e4751f05ed3a659019ee1b/lib/minitest/expectations.rb#L11-L14) you can't use the object monkey patches inside something that is threaded because of the thread local usage.  The `expect` method will allow people to continue to use the expectation syntax but do it in a different thread

### DIFF
--- a/lib/minitest/expectations.rb
+++ b/lib/minitest/expectations.rb
@@ -9,8 +9,9 @@
 #
 #     it "should still work in threads" do
 #       my_threaded_thingy do
-#         (1+1).must_equal 2  # bad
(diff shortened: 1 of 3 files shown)


zenspider:
Done via 9e78cc97

PR #380, opened 2013-12-24
Makes mocks debugger-friendly by exposing instance vars

edward:
When debugging using byebug or a similar debugger, I’d like to be able to inspect the instance variables of a mock. Doing so makes it easy to see what the mock has been set to expect, etc.

This change restores the debugger’s ability to perform that inspection.


zenspider:
I've hand merged this. Thanks!

edward:
Yussss… thanks!

Issue #425, opened 2014-05-02
Feature request: Run suites in random order


judofyr:
I just discovered we had a order-dependent test in our codebase that looked like this:

``` ruby
# in file test_abc.rb (loaded first)
class TestFoobar < Minitest::Test
  def test_foobar
    assert User.find(1) # hardcoded ID
  end
end
```

It would be nice if test shuffling were also applied to suites to catch faulty tests like these:

``` diff
diff --git a/lib/minitest.rb b/lib/minitest.rb
index 106faa6..4893da3 100644
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -138,7 +138,7 @@ module Minitest
   # loaded if a Runnable calls parallelize_me!.

   def self.__run reporter, options
-    suites = Runnable.runnables
+    suites = Runnable.runnables.shuffle
     parallel, serial = suites.partition { |s| s.test_order == :parallel }

     # If we run the parallel tests before the serial tests, the parallel tests
```

PR #426, opened 2014-05-03
Run test suites in random order:

judofyr:
Randomization helps detect order-dependent tests (aka. crappy tests). Test
_methods_ have been randomized for a long time; this commit randomizes the
order test _suites_ are run in as well.

Example:

```
class TestFoo < Minitest::Test; end
class TestBar < Minitest::Test; end
```

Before: TestFoo would always run before TestBar. Tests within TestFoo/TestBar
would be randomized, but the suite themselves were called in a specific order.

After: TestBar might (or might) not run before TestFoo. It's currently not
possible to disable suite randomization, but if you have order-dependent test
you really ought to place them in the same class and use
i_suck_and_my_tests_are_order_dependent!

(This pull request implements #425)

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -138,7 +138,7 @@ module Minitest
   # loaded if a Runnable calls parallelize_me!.
 
   def self.__run reporter, options
-    suites = Runnable.runnables
+    suites = Runnable.runnables.shuffle
     parallel, serial = suites.partition { |s| s.test_order == :parallel }
 
     # If we run the parallel tests before the serial tests, the parallel tests


sagotsky:
I know this was rejected a while ago, but I'd like to make a case for it.  

I have flaky tests in my test suite.  They don't flake out when run in isolation, so I'm trying to see if they're order dependent.  I'd really like to be able to run these files in a specific order to confirm and eventually fix that.  While I generally agree with minitest's opinion on order dependent tests, it's blocking me from fixing my order dependent tests.

blowmage:
@sagotsky Take a look at [minitest-bisect](https://github.com/seattlerb/minitest-bisect). It runs your r tests over and over to find those kinds of problems.

zenspider:
> P.S. I expect someone to bitch about this change at some point.

I told ya so! 😛

@sagotsky this is _exactly_ what minitest-bisect is for. Let me know if you need any help with it (by opening an issue on minitest-bisect).

sagotsky:
Whoa, that does look perfect for this.  Now the trick is finding a seed...

PR #429, opened 2014-05-10
Fix at_exit in fork

ioquatix:
MiniTest `autorun` currently fails to work correctly when using `Process.fork` in a test. This patch demonstrates the issue and proposes a fix.

### DIFF
--- a/test/minitest/test_minitest_fork.rb
+++ b/test/minitest/test_minitest_fork.rb
@@ -1,6 +1,16 @@
 require 'minitest/autorun'
 
 class TestMinitestFork < Minitest::Test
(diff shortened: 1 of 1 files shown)


tonytonyjan:
closed without a reason?

ioquatix:
@tonytonyjan I made this PR to fix the problem but it wasn't accepted because @zenspider doesn't understand why this is a problem or why it should be fixed. Honestly, I find the attitude frustrating and confusing, since it's clearly a mis-interpretation of the specs. I moved to RSpec and have never looked back. RSpec 3 is awesome and doesn't have any issues like this.

PR #453, opened 2014-08-25
Let CompositeReporter act like any other Reporter

teoljungberg:
So the reporters can share all the same attributes, like `io-type

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -582,15 +582,15 @@ module Minitest
   ##
   # Dispatch to multiple reporters as one.
 
-  class CompositeReporter < AbstractReporter
+  class CompositeReporter < Reporter
     ##
     # The list of reporters to dispatch to.
 
     attr_accessor :reporters
 
-    def initialize *reporters # :nodoc:
-      super()
-      self.reporters = reporters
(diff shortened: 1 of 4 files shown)


zenspider:
I've poked at this a number of times since it was first submitted. From the beginning I've had a gut feeling about not liking it but not a good way to put that into words that makes sense to others. It centers around the Composite pattern but is so deeply ingrained that it is more a visceral feeling than a clean intellectual description of what I don't like.

I've found a middle ground that allows me to use about half your patch and cleans up the awkward code without making me feel anything in my gut. 

Thank you for pointing it out. I'm sorry I can't provide better feedback. Please don't let this dissuade you from further PRs.

teoljungberg:
I see. Let me know if you figure out any concrete feedback.

zenspider:
> On Jan 26, 2016, at 00:19, Teo Ljungberg <notifications@github.com> wrote:
> 
> I see. Let me know if you figure out any concrete feedback.

Is my change acceptable to you?

teoljungberg:
> Is my change acceptable to you?

It makes sense.

I was more interested in why the `CompositeReporter` doesn't inherit from `Reporter`, since it share similar traits.

PR #457, opened 2014-09-04
Adding "remove_expectations" method to Minitest::Mock

ggarnier:
I've added a new method to remove all expectations on a specific method:

    @mock.expect(:meaning_of_life, 42)
    @mock.meaning_of_life # => 42

    @mock.remove_expectations(:meaning_of_life)
    @mock.meaning_of_life # raises MockExpectationError

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -77,6 +77,19 @@ module Minitest # :nodoc:
       self
     end
 
+    ##
+    # Removes all expectations from a method +name+.
+    #
+    #   @mock.expect(:meaning_of_life, 42)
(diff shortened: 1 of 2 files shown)


zenspider:
No. Sorry. That's incredibly unclear and I don't want to support it. It honestly reads worse in your second version. If you want to clean things up, consider adding a helper method:

```ruby
def expect_logger val = String
  logger.expect(:info, nil, [val])
end
```

But I can't support remove_expectations. Sorry. You're welcome to make it a plugin gem or keep it as a monkeypatch in your helper, but I seriously recommend you don't test that way. It'll just be confusing to you in 3 months.

PR #472, opened 2014-11-17
README is rdoc now

mjacobus:
Hi guys. The README is written in the rdoc format but it was named as a txt file. Here's a little change to make it more appealing as it was meant to be.


claudiug:
:+1:

blowmage:
You need to specify the readme in the Rakefile for this change to work. See [how minitest-rails does it](https://github.com/blowmage/minitest-rails/blob/master/Rakefile#L17).

PR #480, opened 2014-12-10
Randomization stuff

deivid-rodriguez:
Hi!

I was reading through minitest code and I found one place which didn't seem right: https://github.com/seattlerb/minitest/blob/master/lib/minitest/test.rb#L67. Why are we sorting the array before shuffling it? I thought that was not neccessary and removing it would save as some CPU cycles, so I removed it. Then the test suite start failing and reading through it I found this: https://github.com/seattlerb/minitest/blob/master/test/minitest/metametameta.rb#L86. I'd say the only reason both lines are there is to make the tests pass... so I've removed them if this PR.

Let me know whether all this make any sense.

Cheers!

### DIFF
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -64,7 +64,7 @@ module Minitest
       case self.test_order
       when :random, :parallel then
         max = methods.size
-        methods.sort.sort_by { rand max }
+        methods.sort_by { rand max }
       when :alpha, :sorted then
         methods.sort
       else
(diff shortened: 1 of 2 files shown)


zenspider:
Sorry, but I'm uninterested in this patch. There's nothing in ruby's definition that specifies what order `#methods` should return things, so it is implementation dependent. Having a single extra `#sort` call is not detrimental to the readability.

PR #486, opened 2015-01-20
Avoid eagerly starting threads.

dylanahsmith:
cc @camilo

## Problem

Commit 34760e3b268bc1bb4ac5fe1a44ef1ff0a2f9bd4d made it such that `require 'minitest'` starts 2 threads by default.

This can easily result in minitest threads running outside of the test environment (even in production).  For instance, activesupport has a [runtime dependancy on minitest](https://github.com/rails/rails/blob/v4.1.9/activesupport/activesupport.gemspec#L26), then using a gem like prototype-rails that [extends a test case class](https://github.com/rails/prototype-rails/blob/v4.0.0/lib/prototype-rails/on_load_action_view.rb#L17) will result in two minitest threads starting in any environment.

## Solution

Use a memoized accessor method to get the `Parallel::Executor` instance so that the threads are only started when there are tests to run.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -19,8 +19,10 @@ module Minitest
   ##
   # Parallel test executor
 
-  mc.send :attr_accessor, :parallel_executor
-  self.parallel_executor = Parallel::Executor.new((ENV['N'] || 2).to_i)
+  mc.send :attr_writer, :parallel_executor
+  def self.parallel_executor
+    @parallel_executor ||= Parallel::Executor.new((ENV['N'] || 2).to_i)
+  end
 
   ##
   # Filter object for backtraces.


zenspider:
How is this a problem, esp in production?

dylanahsmith:
Ideally, in production, I don't want any test framework running at all.

Having unnecessary threads running is a waste the system's resources.  It is also very confusing (and scary) when debugging a production issue (e.g. deadlocked job) to see test code running.

zenspider:
How/why are you seeing minitest running in production? _You_ have to load the code for it to run, so I'm confused.

blowmage:
I would bet you that this is due to Bundler's auto-require. A workaround would be to specify the following in the rails app's Gemfile:

```ruby
gem "minitest", require: false
```

dylanahsmith:
> I would bet you that this is due to Bundler's auto-require.

Nope, it is happening because of rails auto-loading.

To reproduce this, create a rails app

```
$ rails new testapp
$ cd testapp
$ bundle exec rails console
irb> defined?(Minitest)
=> nil
irb> ActionView::TestCase
irb> defined?(Minitest)
=> "constant"
irb> Thread.list.each{ |t| puts t.backtrace; puts }
...

/Users/dylansmith/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/minitest-5.5.1/lib/minitest/parallel.rb:23:in `pop'
/Users/dylansmith/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/minitest-5.5.1/lib/minitest/parallel.rb:23:in `block (2 levels) in initialize'

/Users/dylansmith/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/minitest-5.5.1/lib/minitest/parallel.rb:23:in `pop'
/Users/dylansmith/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/minitest-5.5.1/lib/minitest/parallel.rb:23:in `block (2 levels) in initialize'
```

That `ActionView::TestCase` line is what prototype-rails is doing on this line: https://github.com/rails/prototype-rails/blob/v4.0.0/lib/prototype-rails/on_load_action_view.rb#L17

The point is that it is very easy to accidentally autoload minitest, and that is all it takes to start those other threads.

blowmage:
Seems like prototype-rails is the bad actor then. Modifying ActionView::TestCase should probably be wrapped with a `if ::Rails.env.test?` condition.

lib/minitest.rb
```diff
@@ -19,8 +19,9 @@ module Minitest
   ##
   # Parallel test executor
 
-  mc.send :attr_accessor, :parallel_executor
-  self.parallel_executor = Parallel::Executor.new((ENV['N'] || 2).to_i)
+  def self.parallel_executor
```
blowmage: I don't mind memorizing the attribute, I usually prefer it. But we still need to be able to set the attribute even when memorizing it.

```ruby
mc.send :attr_writer, :parallel_executor
def self.parallel_executor
  @parallel_executor ||= Parallel::Executor.new((ENV['N'] || 2).to_i)
end
```

dylanahsmith:
Yes, that is definitely a bug in prototype-rails, which should get fixed by pull request https://github.com/rails/prototype-rails/pull/33.  However, just fixing the problem there is a brittle solution, since the fact remains that is is easy to have some gem provide some TestCase extension which autoloads minitest, in which case it doesn't make sense for it to start threads automatically.

lib/minitest.rb
dylanahsmith: Sure, I've added the `mc.send :attr_writer, :parallel_executor` line

tenderlove:
> Having unnecessary threads running is a waste the system's resources.

How much of a waste is it?  They never wake up unless you're running tests.

I don't really mind if we add this code, but this patch makes `self.parallel_executor` not thread safe.  Given that the method returns an object for coordinating threads, that makes me cringe a bit.

If we could quantify how much of a waste it is, then I'd feel more happy about this change.  Besides possibly being a waste, I'm still not clear why we need this change.

zenspider:
Yeah. I understand the scenario better, but I still don't see the concern/priority. Someone/something errantly loading `ActiveSupport::TestCase` in production is just wrong. I'd be much more concerned about that (and other magical features doing other things you don't know about until you have a problem).

As you can see in the backtraces you printed, the threads are sleeping on `Queue#pop`. It's not really something to be concerned about wrt resources. It _is_ something to be concerned about in terms of "what's going on underneath the covers".

tenderlove:
We *could* do a patch like this:

```diff
diff --git a/lib/minitest.rb b/lib/minitest.rb
index b2c5cad..f5e2085 100644
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -123,6 +123,7 @@ module Minitest
     self.init_plugins options
     self.reporter = nil # runnables shouldn't depend on the reporter, ever
 
+    self.parallel_executor.start
     reporter.start
     __run reporter, options
     self.parallel_executor.shutdown
diff --git a/lib/minitest/parallel.rb b/lib/minitest/parallel.rb
index 107d158..9cd76e1 100644
--- a/lib/minitest/parallel.rb
+++ b/lib/minitest/parallel.rb
@@ -17,6 +17,13 @@ module Minitest
       def initialize size
         @size  = size
         @queue = Queue.new
+        @pool  = nil
+      end
+
+      ##
+      # Start the executor
+
+      def start
         @pool  = size.times.map {
           Thread.new(@queue) do |queue|
             Thread.current.abort_on_exception = true
```

That would give folks who implement parallel runners a hook point to allocate resources, maintain thread safety, and fix @dylanahsmith's concerns. (I personally would like a start hook for a forking runner)

dylanahsmith:
@tenderlove :+1: to your proposed patch.

Closing since my code is indeed not thread-safe.

PR #487, opened 2015-01-20
Add minitest.gemspec and Gemfile

dylanahsmith:
This will make it easier to contribute to minitest, since a developer can just do `bundle install` to install the dependancies.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+Gemfile.lock
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,2 @@
+source 'https://rubygems.org'
+gemspec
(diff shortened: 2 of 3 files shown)


zenspider:
No thanks, for many of the same reasons listed here: https://github.com/sparklemotion/nokogiri/blob/master/Y_U_NO_GEMSPEC.md

dylanahsmith:
> No thanks, for many of the same reasons listed here: https://github.com/sparklemotion/nokogiri/blob/master/Y_U_NO_GEMSPEC.md

The concern there seems to be mostly around native extensions.  C code is prone to memory corruption, leaking memory, crashes and remote code execution vulnerabilities.  In contrast, minitest is a pure ruby test framework, so is not prone to these problems.

zenspider:
C code is just one aspect. Stability is another. I'm sticking with my Hoe setup.

PR #488, opened 2015-01-24
Print command that runs a single test for each of the failed tests

jirutka:
Make @tenderlove (and his cat) happier with the Minitest. :smiley_cat:

See section “Things I dislike about Minitest” of the blog post [My experience with Minitest and RSpec](http://tenderlovemaking.com/2015/01/23/my-experience-with-minitest-and-rspec.html).

Example:

```
...
285 runs, 841 assertions, 2 failures, 0 errors, 0 skips

# Failed tests:

ruby test/minitest/test_minitest_unit.rb -n 'TestMinitestRunner#test_run_error_teardown'
ruby test/minitest/test_minitest_reporter.rb -n 'TestMinitestReporter#test_start'
```

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -2,6 +2,7 @@ require "optparse"
 require "thread"
 require "mutex_m"
 require "minitest/parallel"
+require "pathname"
 
 ##
 # :include: README.txt
@@ -560,6 +561,7 @@ module Minitest
       io.puts statistics
       io.puts aggregated_results
       io.puts summary
+      io.puts failed_tests unless passed?
     end
 
     def statistics # :nodoc:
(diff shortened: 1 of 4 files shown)


lib/minitest.rb
```diff
@@ -592,6 +594,23 @@ def summary # :nodoc:
...
+    def failed_tests # :nodoc:
+      lines = [ "\n# Failed tests:\n" ]
+
+      results.reject(&:skipped?).each do |result|
+        file_path = relative_from_pwd(result.failure.location.sub(/:\d+$/, ""))
+        lines << "ruby #{file_path} -n #{result.class}##{result.name}"
+      end
+
+      lines.join("\n") + "\n"
+    end
```
jirutka: Note: This can be also written more functionally like:

```ruby
    def failed_tests
      results.reject(&:skipped?).map { |result|
        file_path = relative_from_pwd(result.failure.location.sub(/:\d+$/, ""))
        "ruby #{file_path} -n #{result.class}##{result.name}"
      }.unshift("\n# Failed tests:\n").join("\n") + "\n"
    end
```

…but it’s IMHO less readable.

zenspider:
I don't want this code in minitest. I'm working on an alternative. Thanks tho.

PR #489, opened 2015-01-27
Clarify in readme that minitest/XXX_plugin.rb must be under directory on load path

garysweaver:
Was working on a project in vanilla Ruby, was trying to load a minitest plugin from `test/minitest/some_plugin.rb`, but had forgotten to put test directory on load path, e.g. in Rakefile via:

```ruby
$LOAD_PATH.unshift File.join(File.expand_path(File.dirname(__FILE__)), 'test')
```

So, adding a slight reminder in the readme to help remember load path.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -265,13 +265,14 @@ provided via plugins. To see them, simply run with `--help`:
 
 == Writing Extensions
 
-To define a plugin, add a file named minitest/XXX_plugin.rb to your
+To define a plugin, add a file called XXX_plugin.rb into a directory
+called minitest under a directory that is on the load path of your
 project/gem. Minitest will find and require that file using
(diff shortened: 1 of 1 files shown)


zenspider:
I didn't like your wording. So I set it back to the previous and added "That file must be discoverable via ruby's LOAD_PATH (via rubygems or otherwise)." after the first sentence.

As a sidenote: I really don't recommend putting minitest plugins in your test directory, except maybe while it is under development. But yeah, the hoe's test task always adds test to the load path, so I don't hit snafus like this. If you're using rake's test task, you only need to push test onto libs:

```ruby
Rake::TestTask.new do |t|
  t.libs << "test"
  # ...
end
```

zenspider:
submitted.

garysweaver:
@zenspider Perfect, thanks!

PR #491, opened 2015-01-27
lazily start the parallel executor

tenderlove:
This has two benefits:

* Gives a hook for exetutors to start threads / processes
* Avoids booting threads when minitest is loaded but not in use


zenspider:
merged. Thanks!

PR #496, opened 2015-02-05
Include minitest-profile as a known extension

nmeans:


zenspider:
Done. Sorry I took so long!

PR #497, opened 2015-02-05
Move parallel executor management to .__run

elia:
Seems a better place for managing parallel runs.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -126,7 +126,6 @@ module Minitest
     self.parallel_executor.start if parallel_executor.respond_to?(:start)
     reporter.start
     __run reporter, options
-    self.parallel_executor.shutdown
     reporter.report
 
     reporter.passed?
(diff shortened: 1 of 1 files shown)


zenspider:
This doesn't look right. It introduces a dissymmetry between the starting of the parallel runner and the shutdown of it. `__run` is in place in order to make it overridable and you're now pushing a responsibility down to anyone who overrides it that they're responsible for cleanup.

@tenderlove opinions?

tenderlove:
I don't get it. @elia why is this better?

elia:
__run is where parallel tests are started and seemed a better place to shutdown them. Additionally because I didn't see any other reference to parallel inside self.run. 


As of subclassing I see it's marked as internal with a possibly out of sync comment about parallel_each.rb. My initial intent anyway was to extract parallel stuff and keep it all in the same place (now spans through minitest.rb, my/parallel.rb and mt/test.rb).

elia:
Forgive me, even if I looked at it many times I totally missed the parallel_executor.start

😓

Maybe I shouldn't open pull-requests after 1am

PR #501, opened 2015-02-09
Provide stub explanation and syntax

kkirsche:
Provide a clearer of what stubs are and the syntax for how to use them.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -216,7 +216,7 @@ Output is tab-delimited to make it easy to paste into a spreadsheet.
 
 Stubs simulate the behaviors of software components (or modules) that a module undergoing tests depends on. Stubs are defined in the following way:
 
-    Klass.stub :method, return_value, block_argument do   # stub goes away once the block is done
+    Klass_Or_KlassInstance.stub :method, return_value, block_argument do   # stub goes away once the block is done
       # Assertions go here
     end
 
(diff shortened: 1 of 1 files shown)


README.txt
```diff
@@ -214,6 +214,14 @@ Output is tab-delimited to make it easy to paste into a spreadsheet.
 
 === Stubs
 
+Stubs simulate the behaviors of software components (or modules) that a module undergoing tests depends on. Stubs are defined in the following way:
+
+    Klass.stub :method, return_value, block_argument do   # stub goes away once the block is done
```
zenspider: But, stubs aren't class specific. They can be run on any instance, as demonstrated in our previous dialog.

kkirsche:
How so? In the example provided we aren't using the standard stubs, we are simply overriding the method within the subclass so it doesn't call the main class's method.

kkirsche:
Fixed. Sorry about that. Missed the stub with the `StringIO` instance being yielded and saw first just the class override.

kkirsche:
Wanted to follow up and see what your thoughts were of the fix or if more was needed. Thanks again for your time and assistance reviewing this

kkirsche:
Wanted to follow up on this as I haven't heard back from @zenspider

zenspider:
I'm just not sure that this really clarifies anything. "Stubs simulate the behaviors of software components" certainly feels wrong to me. I'll poke at this.

zenspider:
I've submitted a change that I hope clarifies things better.

PR #507, opened 2015-02-26
stop allocating time objects

tenderlove:
Instead of allocating time objects, we can use the monotonic clock (if
it is available).  This gives us the benefit of dropping allocations and
will avoid incorrect time reporting due to the system clock changing or
people stubbing Time.now and they forget to unstub (lol).

Here is the test I used:

```ruby
require 'minitest'
require 'allocation_tracer'

class AmericaTest < Minitest::Test
  def test_freedom
    assert true
  end
end

result = ObjectSpace::AllocationTracer.trace do
  3000.times do
    Minitest.run_one_method(AmericaTest, 'test_freedom')
  end
end

p ObjectSpace::AllocationTracer.allocated_count_table[:T_DATA]
```

Before this patch, each test would allocate 9 T_DATA objects per test.
After this patch, it allocates 6 T_DATA objects per test.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -507,7 +507,7 @@ module Minitest
     end
 
     def start # :nodoc:
-      self.start_time = Time.now
+      self.start_time = Minitest.clock_time
     end
 
     def record result # :nodoc:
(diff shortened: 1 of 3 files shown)


zenspider:
Merged. Thanks!

PR #511, opened 2015-03-13
Add minitest-matchers_vaccine to known extensions

rmm5t:


zenspider:
I dunno what it is, but it sounds awesome. Thanks!

PR #512, opened 2015-03-15
Fix failing test under Windows

kimhmadsen:
Windows echo just echos the apostrophes therefore this little hack.


zenspider:
Done. Thanks!

PR #525, opened 2015-04-20
Add minitest-rerun-options to list of extensions

ivantsepp:
Adds `minitest-rerun-options` to list of extensions. 

I initially did some searching and couldn't find what I was looking for. So I worked on this and then when I finished, I decided to browse some more and found that https://github.com/grosser/minitest-rerun and https://github.com/seattlerb/minitest-sprint already does something similar (*face palm*).

Anyways do you still think there's value in this extension? I like the option for rerunning all failed tests. Is this something I should add to `minitest-sprint`?

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -514,6 +514,8 @@ minitest-profile            :: List the 10 slowest tests in your suite.
 minitest-rails              :: Minitest integration for Rails 3.x.
 minitest-rails-capybara     :: Capybara integration for Minitest::Rails.
 minitest-reporters          :: Create customizable Minitest output formats.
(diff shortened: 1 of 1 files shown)


zenspider:
Not against it, but you might want to look at minitest-autotest.

ivantsepp:
`minitest-autotest` is awesome! I was trying to figure out how to rerun the last command with failing tests but it seems that `minitest-autotest` handles that. Thanks for this! 

My use case is to run failing tests reported from CI so being able to copy/paste commands to my local machine is really helpful. 

Thanks again for looking into this!

PR #528, opened 2015-04-22
Add minitest-junit

aespinosa:
I added some Jenkins-specific quirks when interprets the junit files.

https://github.com/aespinosa/minitest-junit


zenspider:
done. Thanks!

PR #529, opened 2015-04-26
Add minitest-fail-fast to README

teoljungberg:
It was mentioned/asked for during your talk at Railsconf about the fail-fast feature for rspec, and the implementation that was out there no longer worked with Minitest 5. So I hacked this together on the flight back home, I hope that solves the problems for who-ever wants this feature

Hence, I added it to the README.

Source code : https://github.com/teoljungberg/minitest-fail-fast


teoljungberg:
Any ideas on this @zenspider ?

zenspider:
done. thanks!

PR #532, opened 2015-04-29
Fix usage of proc targets in expectations for assertions that don't take blocks

jeremyevans:
Previously, an expectation like this would raise an ArgumentError

  proc{}.wont_equal nil

Only expectations for assertions that explicitly use pass blocks
should treat proc targets specially.  In all other cases, the proc
target should be treated as a regular argument.

While here, use __FILE__ and __LINE__ + 1 on the class_eval for
better backtraces.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -2,8 +2,15 @@ require "minitest/test"
 
 class Module # :nodoc:
   def infect_an_assertion meth, new_name, dont_flip = false # :nodoc:
+    if dont_flip == :block
+      block = true
+      dont_flip = false
+    else
+      block = false
+    end
+
     # warn "%-22p -> %p %p" % [meth, new_name, dont_flip]
-    self.class_eval <<-EOM
+    self.class_eval <<-EOM, __FILE__, __LINE__ + 1
       def #{new_name} *args
         Minitest::Expectation.new(self, Minitest::Spec.current).#{new_name}(*args)
       end
@@ -14,7 +21,7 @@ class Module # :nodoc:
         case
         when #{!!dont_flip} then
           ctx.#{meth}(target, *args)
-        when Proc === target then
+        when #{block} && Proc === target then
           ctx.#{meth}(*args, &target)
         else
           ctx.#{meth}(args.first, target, *args[1..-1])
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -2,12 +2,8 @@ require "minitest/test"
 
 class Module # :nodoc:
   def infect_an_assertion meth, new_name, dont_flip = false # :nodoc:
-    if dont_flip == :block
-      block = true
-      dont_flip = false
-    else
-      block = false
-    end
+    block = dont_flip == :block
+    dont_flip = false if block
 
     # warn "%-22p -> %p %p" % [meth, new_name, dont_flip]
     self.class_eval <<-EOM, __FILE__, __LINE__ + 1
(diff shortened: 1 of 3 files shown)


lib/minitest/spec.rb
```diff
@@ -2,8 +2,15 @@
 
 class Module # :nodoc:
   def infect_an_assertion meth, new_name, dont_flip = false # :nodoc:
+    if dont_flip == :block
+      block = true
+      dont_flip = false
+    else
+      block = false
+    end
```
zenspider:
```ruby
block = dont_flip == :block
dont_flip = false if block
```

test/minitest/test_minitest_spec.rb
```diff
@@ -291,6 +302,7 @@ def assert_triggered expected = "blah", klass = Minitest::Assertion
     assert_triggered "msg.\nExpected 1 to not be equal to 1." do
       1.wont_equal 1, "msg"
     end
+
```
zenspider: nix

zenspider:
Sorry, it's late and I'm sure I'm being a bit obtuse, but what is the use case here?

jeremyevans:
This bug caused quite a few errors when converting Roda's tests from rspec to minitest. In one of the cases, I'm checking that some value is not nil:

```ruby
object.method.must_equal nil
object.other_method
object.method.wont_equal nil
```

The `method` currently return a proc, but in the future that could change to something else.

blowmage:
FWIW, one way to solve this in your tests is to change the expectation to the following:

```ruby
object.method.wont_be :nil?
```

This doesn't correct the underlying issue, but it will unblock you for the time being. I have no position on whether this should be changed or not.

jeremyevans:
I've been converting most of my libraries from rspec to minitest/spec over the last week, and I've had to work around this bug in multiple places.  Here's my favorite example of it:

    proc{}.must_be_kind_of(Proc)

There's not really a good way to express that with the spec syntax due to this bug, you basically have to switch to:

    assert_kind_of Proc, proc{}

jeremyevans:
@zenspider Have you given this some more thought?  This fixes a bug in minitest, and the only real way to work around it is to use an assertion instead of an expectation.  I currently work around this bug 7 times in Sequel's specs:

```
spec/core/connection_pool_spec.rb:    assert_equal pr, p.after_connect
spec/core/dataset_spec.rb:    assert_equal @dataset.row_proc, clone.row_proc
spec/core/schema_generator_spec.rb:    assert_kind_of Proc, @constraints[1][:check]
spec/extensions/composition_spec.rb:    assert_kind_of Proc, r[:composer]
spec/extensions/composition_spec.rb:    assert_kind_of Proc, r[:decomposer]
spec/model/plugins_spec.rb:    assert_equal b, m.block
spec/model/plugins_spec.rb:    assert_equal b, m.block2
```

and twice in Roda's specs:

```
spec/plugin/path_spec.rb:    assert_equal @app.path_block(c), b
spec/integration_spec.rb:    assert_kind_of Proc, app.app
```

zenspider:
The fact that you can't currently call any of the non-proc-oriented expectations with a proc on LHS means that this code blows up:

```ruby
proc { 42 }.must_equal 42
```

with an argument error. This is good and means that nobody is writing code that expects procs on LHS to be called (except in the proc-oriented cases, obvs). Changing this behavior seems fine then.

zenspider:
I tweaked a bit (mostly: you and I test differently). Essence is the same. Thanks!

PR #533, opened 2015-04-29
Make assert_raises respect subclasses

jeremyevans:
Having assert_raises not respect subclasses leads to fragile specs,
where you cannot change your code to raise a subclass of an
existing exception class without breaking your specs.

assert_raises should operate similar to a rescue block.  If you do

```ruby
  assert_raises StandardError do
    raise RuntimeError
  end
```

that should not raise an assertion error, because RuntimeError is
a subclass of StandardError and if you do:

```ruby
  begin
    raise RuntimeError
  rescue StandardError
  end
```

the RuntimeError will be rescued.

This breaks backward compatibility, but IMO is the correct
behavior.  If you truly want to check for a specific exception
class:

```ruby
  e = assert_raises StandardError do
    raise RuntimeError
  end
  assert_instance_of e, StandardError
```

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -298,11 +298,7 @@ module Minitest
         raise e
       rescue Exception => e
         expected = exp.any? { |ex|
-          if ex.instance_of? Module then
-            e.kind_of? ex
-          else
-            e.instance_of? ex
-          end
+          e.kind_of? ex
         }
 
         assert expected, proc {
--- a/test/minitest/test_minitest_unit.rb
+++ b/test/minitest/test_minitest_unit.rb
@@ -1236,26 +1236,10 @@ class TestMinitestUnitTestCase < Minitest::Test
     assert_equal expected, e.message
   end
 
-  def test_assert_raises_triggered_subclass
-    e = assert_raises Minitest::Assertion do
-      @tc.assert_raises StandardError do
-        raise AnError
-      end
+  def test_assert_raises_subclass
+    @tc.assert_raises StandardError do
+      raise AnError
     end
-
-    expected = clean <<-EOM.chomp
-      [StandardError] exception expected, not
-      Class: <AnError>
-      Message: <\"AnError\">
-      ---Backtrace---
-      FILE:LINE:in \`test_assert_raises_triggered_subclass\'
-      ---------------
-    EOM
-
-    actual = e.message.gsub(/^.+:\d+/, "FILE:LINE")
-    actual.gsub!(/block \(\d+ levels\) in /, "") if RUBY_VERSION >= "1.9.0"
-
-    assert_equal expected, actual
   end
 
   def test_assert_respond_to


zenspider:
Not sure which is worse: "fragile specs" or "vague(*) specs".

*) "weak" or "loose" might be a better word. Not sure.

jeremyevans:
I think fragile specs are worse, since they can result in braking specs when a more generic exception class is replaced with a subclass that is more specific.  With vague/weak/loose specs, they are accurate, just not perfectly precise.

A spec that says something must raise `FooError` tells the user that if they rescue `FooError`, the exception will be handled.  If the library changes to raise a subclass of `FooError` named `FooBarError`, this should be completely backwards compatible and not break any specs.  If the library maintainer wants to change the specification to state that now `FooBarError` is raised, they can do that, but it shouldn't be forced on them by the testing library.

Basically, minitest's current behavior means you can't make backwards compatible changes to your exception class hierarchies without breaking your specs, which I think is a bad idea.  Backwards compatible changes should not break specs.

jeremyevans:
Here's another example of where the current behavior bites you:

```ruby
require 'minitest/autorun'

describe 'appending to a frozen string' do
  it "raises an exception for invalid input" do
    proc{'1'.freeze << 'a'}.must_raise # ???
  end
end
```

What do you put there?  You could put `RuntimeError`, but that doesn't work on ruby 1.8.  You could put `TypeError`, but that doesn't work on ruby 1.9+.  If you want your specs to run on both 1.8 and 1.9+, you have to go out of your way to get this to work (`FrozenError = RUBY_VERSION >= '1.9' ? RuntimeError : TypeError`), unless you use this patch, in which case you can use `StandardError` (or no arguments if you also use the patch in #534).

As I'm writing specs for a library with a lot of frozen objects (Roda), and which runs on ruby 1.8.7 to 2.2, this issue is very prevalent.  In the specs I'm writing, I really don't care what exception class ruby is raising since I don't have any control over it.  I just want to spec that it is raising an exception if you do something you aren't supposed to after freezing.

jeremyevans:
When converting Sequel to use minitest, this issue caused some additional issues in the adapter used.  Many times in Sequel's specs, I would be using something similar to:

```
proc{}.must_raise Sequel::DatabaseError
```

This would fail depending on the adapter used, with some adapters raising more specific exception classes.  In some cases I have to include 3 separate arguments, each a more specific subclass of the previous class:

```
proc{}.must_raise Sequel::DatabaseError, Sequel::ConstraintViolation, Sequel::CheckConstraintViolation
```

While I don't think the current behavior could be described as a bug, I do think it violates the principle of least surprise.

zenspider:
@tenderlove and I talked about this over lunch the other day.

As far as backwards compatibility goes, if your tests are passing, they'll stay passing. If they're failing (because of `assert_raises`) then they _may_ go green (but probably won't, because why would you leave such a trivial failure in place?). There's no situation I can come up with where your green tests will go red.

I added a negative test and collapsed the impl to a 1 liner, but is otherwise the same.

PR #534, opened 2015-04-29
Make assert_raises with no arguments assume StandardError

jeremyevans:
This reflects ruby's behavior, where rescue with no exception
classes assumes StandardError.


zenspider:
I did this a tad differently, but the essence is the same. Thanks!

PR #538, opened 2015-05-02
Adds optional delegation mechanism to extend object with a mock

zamith:
I've come by certain situations where I want to mock a specific method
on an object, and verify that it is called, but as far as the other
methods go, I want them to do their normal job.

Without this delegator, I would have to mock every single method that
gets called on the object the mock is mocking, which is both tedious,
error prone and leads to a path of brittle tests.

With the delegator I can mock "on top" of this object, allowing me to
mock a specific method, making my tests more intent revealing and
smaller.

This has costs, albeit small, on performance, but they won't get
triggered unless you opt in to the functionality. It might also be
argued that it has the drawback of hiding some implicit dependencies. I
believe, however, that it is a net win.

Am I missing some other way of doing this? Is this a terrible solution
for some reason I cannot see at the moment? @zenspider, what do you
think?

PS: This is a draft of the solution, I can add tests for it if you like it.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -36,7 +36,8 @@ module Minitest # :nodoc:
       end
     end
 
-    def initialize # :nodoc:
+    def initialize(delegator = nil) # :nodoc:
+      @delegator = delegator
       @expected_calls = Hash.new { |calls, name| calls[name] = [] }
       @actual_calls   = Hash.new { |calls, name| calls[name] = [] }
     end
@@ -113,8 +114,11 @@ module Minitest # :nodoc:
 
     def method_missing(sym, *args, &block) # :nodoc:
       unless @expected_calls.key?(sym) then
-        raise NoMethodError, "unmocked method %p, expected one of %p" %
-          [sym, @expected_calls.keys.sort_by(&:to_s)]
+        if @delegator
+          delegate_method(sym, *args, &block)
+        else
+          raise_unmocked_method_error(sym)
+        end
       end
 
       index = @actual_calls[sym].length
@@ -162,7 +166,21 @@ module Minitest # :nodoc:
 
     def respond_to?(sym, include_private = false) # :nodoc:
       return true if @expected_calls.key? sym.to_sym
+      return true if @delegator && @delegator.respond_to?(sym, include_private)
       __respond_to?(sym, include_private)
+    end
+
+    private
+
+    def delegate_method(sym, *args, &block)
+      @delegator.public_send(sym, *args, &block)
+    rescue NoMethodError
+      raise_unmocked_method_error(sym)
+    end
+
+    def raise_unmocked_method_error(method_name)
+      raise NoMethodError, "unmocked method %p, expected one of %p" %
+        [method_name, @expected_calls.keys.sort_by(&:to_s)]
     end
   end
 end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -114,10 +114,11 @@ module Minitest # :nodoc:
 
     def method_missing(sym, *args, &block) # :nodoc:
       unless @expected_calls.key?(sym) then
-        if @delegator
-          delegate_method(sym, *args, &block)
+        if @delegator && @delegator.respond_to?(sym)
+          return @delegator.public_send(sym, *args, &block)
         else
-          raise_unmocked_method_error(sym)
+          raise NoMethodError, "unmocked method %p, expected one of %p" %
+            [sym, @expected_calls.keys.sort_by(&:to_s)]
         end
       end
 
@@ -168,19 +169,6 @@ module Minitest # :nodoc:
       return true if @expected_calls.key? sym.to_sym
       return true if @delegator && @delegator.respond_to?(sym, include_private)
       __respond_to?(sym, include_private)
-    end
-
-    private
-
-    def delegate_method(sym, *args, &block)
-      @delegator.public_send(sym, *args, &block)
-    rescue NoMethodError
-      raise_unmocked_method_error(sym)
-    end
-
-    def raise_unmocked_method_error(method_name)
-      raise NoMethodError, "unmocked method %p, expected one of %p" %
-        [method_name, @expected_calls.keys.sort_by(&:to_s)]
     end
   end
 end


lib/minitest/mock.rb
```diff
@@ -113,8 +114,11 @@ def verify
 
     def method_missing(sym, *args, &block) # :nodoc:
       unless @expected_calls.key?(sym) then
-        raise NoMethodError, "unmocked method %p, expected one of %p" %
-          [sym, @expected_calls.keys.sort_by(&:to_s)]
+        if @delegator
+          delegate_method(sym, *args, &block)
+        else
+          raise_unmocked_method_error(sym)
+        end
```
zenspider: What does this do? It looks like it calls the method if there is a delegate, and promptly drops the results on the floor and goes on to raise an exception because the method called wasn't mocked. If there is no delegator, it raises an exception. Why?

zenspider:
> This has costs, albeit small, on performance, but they won't get triggered unless you opt in to the functionality.

Don't make claims about performance without a working prototype and actual measurements.

zamith:
@zenspider You're right, I was too broad with that claim. I meant that you won't trigger that code unless you really want to.

Do you think it's worth to benchmark it, or you don't like the idea at all?

zenspider:
I don't think I'm ready to decide. I don't think it currently works.

lib/minitest/mock.rb
zamith: The part where it raises an exception when there is no delegator, was the existing behavior, nothing changed there.

As far as the other branch of the conditional, you're right, it is missing a `return` statement. It is supposed to either return whatever the response is from the delegator or raise if the method is not found there as well (maybe the message could be different in this case).

zamith:
@zenspider It works now, and I've simplified it a bit.

Also, here's an example straight from a refactor to a Rails test. Maybe there is another way around this, and it's always nicer to look at actual code.

```ruby
test "validates_numericality_of for :odd on add message #{name}" do
  Person.validates_numericality_of :title, validation_options.merge(only_integer: true, odd: true)
  @person.title = 0
  errors = Minitest::Mock.new @person.errors
  errors.expect(:add, [], [:title, :odd, validation_options.merge(value: 0)])

  @person.stub(:errors, errors) do
    @person.valid?
  end

  errors.verify
end
```

The `valid?` method will call 2 or 3 more methods on `errors`, which I don't have to worry about with the delegator.

zamith:
@zenspider I've been testing this even more, and I've encountered a situation in which I do care about calls made be a method on the delegator. Here's an example:

```ruby
class A
  def a
    method_i_care_about
  end

  def method_i_care_about
  end
end

test "method I care about is called" do
  a_instance = A.new
  mock = Minitest::Mock.new a_instance
  mock.expect(:method_i_care_about, nil, [])
  a_instance.a
  mock.verify
end
```

This will fail, because even though `a` is called on the mock and then passed down, `method_i_care_about` is called on the actual instance of the delegator, which knows nothing about `@actual_calls`.

I've drafted a quick solution for this, which might look like this:

```ruby
def initialize(delegator = nil) # :nodoc: 
  @expected_calls = Hash.new { |calls, name| calls[name] = [] }
  @actual_calls   = Hash.new { |calls, name| calls[name] = [] }
  Thread.current[:mock_calls] = @actual_calls
  @delegator = delegator
  methods = delegator.methods
  metaclass = class << delegator; self; end
  methods.each do |method_id|
    metaclass.send(:define_method, method_id, proc { |*args, &block| 
      retval = super(*args, &block) 
      Thread.current[:mock_calls][method_id] << {
        :retval => retval,
        :args => args 
      }
      retval
    })
  end
end
```

This is not ideal, but it does fix an issue I believe other people trying to go from other mocking frameworks to this might face.

Also, I think it's worth pointing out at this point that I have not tried to solve this by monkey patching the `Object` class, as most ruby mocking libraries, because I am assuming you had reasons not to do it in the first place.

Please let me know what you think and if this makes any sense or if I'm missing something completely.

zamith:
@zenspider Sorry to bother you, I know you must be busy, but if you could provide any feedback on this, it would be great.

zenspider:
I'm afraid you've pushed into an area that I just can't muster that much concern over as I completely don't test this way. You've put far more effort into minitest/mock.rb over the past 6 months than I have. As a result, you have more of a vested interest in what this should do than I do.

Personally, I don't find these types of tests valuable (I'm a "statist" tester, not a "mockist" tester). I would test it completely differently than you did above and it wouldn't require this mechanism. Here's how I'd do it (and I'm preempting by adding a method called `expensive_call` that is called from `method_i_care_about`:

```ruby
require "minitest/autorun"

class A
  def a
    method_i_care_about
  end

  def method_i_care_about
    expensive_call
  end

  def expensive_call
    sleep 10
  end
end

describe A do
  it "method I care about is called" do
    a = A.new

    a.stub :expensive_call, 42 do
      assert_equal 42, a.a
    end
  end
end
```

So, if the above can make you "see the light", great. If not, I'll cede to your expertise on what minitest/mock should do and merge whatever you want. It sounds like you need to resolve issues caused in your last comment, so let me know if and when to pull the trigger on this PR.

zamith:
I bumped into this while refactoring some rails core tests. I didn't write the tests myself. I was trying to make minitest/mocks behave closer to rspec and mocha.

Let me give another stab at this by trying to test state. I'll let you know once I do.

I don't want to add this overhead, without being completely sure it can really make a difference in certain situations or styles of testing.

zenspider:
Point me at one of the rails tests?

zamith:
Right now the one I'm having trouble with is [this](https://github.com/rails/rails/blob/master/activemodel/test/cases/validations/i18n_validation_test.rb#L60):

```ruby
COMMON_CASES.each do |name, validation_options, generate_message_options|
  test "validates_confirmation_of on generated message #{name}" do
    Person.validates_confirmation_of :title, validation_options
    @person.title_confirmation = 'foo'
    @person.errors.expects(:generate_message).with(:title_confirmation, :confirmation, generate_message_options.merge(attribute: 'Title'))
    @person.valid?
  end
end
```

My implementation with the custom `Minitest::Mock` is:

```ruby
COMMON_CASES.each do |name, validation_options|
  test "validates_confirmation_of on generated message #{name}" do
    Person.validates_confirmation_of :title, validation_options
    @person.title_confirmation = 'foo'
    verify_validity(model: @person, attribute: :title_confirmation, validation: :confirmation, options: validation_options.merge(attribute: 'Title'))
  end
end

def verify_validity(model:, attribute: :title, options: {}, validation:)
  errors = Minitest::Mock.new model.errors
  errors.expect(:add, nil, [attribute, validation, options])

  model.stub(:errors, errors) do
    model.valid?
  end

  errors.verify
end
```

Not sure how to implement this from a statist point of view.

zamith:
@zenspider Any luck with this? If there are no further developments I'd say we could go ahead and merge this, since it's backwards compatible. What do you think?

zenspider:
Sorry. I've had my head in the sand for a couple weeks working on client work. If you want it merged, it gets merged.

zamith:
@zenspider No problem. Yeah, I'd say merge it.

zenspider:
done. thanks and sorry for the delay! bump me on anything else mock related you'd like to push through.

zamith:
Will do. Thanks so much.

PR #541, opened 2015-05-05
Add class method to the overriden methods list

zamith:
Maybe there is some reason not to override this, but I found it useful when running some memory allocation benchmarks.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -12,6 +12,7 @@ module Minitest # :nodoc:
 
     overridden_methods = %w[
       ===
+      class
       inspect
       instance_eval
       instance_variables


zenspider:
Maybe I'm being dense (it is my workout day, use small words): Given how infrequently classes are created, how does this change affect memory profiling?

zamith:
It had no effect on the results of the benchmarks, they just wouldn't run without it. The [allocation_stats](https://github.com/srawlins/allocation_stats) gem calls `class` on the object being profiled and I get an `unmocked method :class` error.

I figured this was a bug of sorts, maybe not. I could also solve it by mocking the `class` method explicitly.

zenspider:
Good enough.

zenspider:
I've got some changes I'm meddling with right now in minitest/mock & tests... once I'm decided and done with that I'll merge this.

zamith:
Ok. Thanks for the time.

zenspider:
Done! This will be released shortly.

PR #543, opened 2015-05-06
Add support for shared descriptions

jeremyevans:
Shared descriptions can be used inside modules that are included
in spec classes.  Usage of modules to share specs is currently
limited to sharing individual specs, shared descriptions allow
you to share spec subclasses across multiple specs.

Currently, if you want to share spec subclasses across multiple
specs, you end up having to do something ugly like:

```ruby
module SharedSpecSubclass
  def self.included(mod)
    super
    pr = proc do
      describe "some description" do
        before do
          # setup
        end

        it "should do something" do end
      end
    end
    if mod.is_a?(Class)
      mod.class_eval(&pr)
    else
      mod.extend(Module.new do
        define_method(:included) do |m|
          super(m)
          m.class_eval(&pr)
        end
      end)
    end
  end
end
```

This commit reduces the above case to:

```ruby
module SharedSpecSubclass
  extend Minitest::Spec::DSL
  shared_description "some description" do
    before do
      # setup
    end

    it "should do something" do end
  end
end
```

The example above is not an artificial example, it is taken
directly from Sequel's specs (which I'm converting from rspec to
minitest/spec).

Keeping with minitest's general approach of "It's Just Ruby",
there is no magic here.  We just save the shared_description
arguments and blocks in an array, and when the class includes
the module with the shared descriptions, we call describe at
that point with the arguments and block.

This supports not only classes including modules with shared
descriptions, but also classes including modules that include
modules with shared descriptions (which Sequel also uses).

Note that this code does not work if you do:

```ruby
  class Module
    include Minitest::Spec::DSL
  end
```

This is because Minitest::Spec::DSL#include will never be called
in this case.  In ruby 2.0+, you can use prepend instead of
include, but I don't think there is a workaround for ruby <2.0.

I'm not tied to this particular API, it was just the simplest
thing I could think of.  I'm fine with an alternative API that
lets you share spec subclasses.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -196,6 +196,84 @@ class Minitest::Spec < Minitest::Test
     end
 
     ##
+    # An array of arguments and blocks to pass to describe for classes that
+    # include the current module.
+
+    def shared_descriptions
+      @shared_descriptions ||= []
+    end
+
+    ##
+    # Add a describe block that will be shared by all classes including the current module.
+    # Example:
+    #
+    #   module EmptyBehavior
+    #     extend Minitest::Spec::DSL
+    #     shared_description "empty" do
+    #       before do
+    #         @that = double
+    #       end
+    #
+    #       it "must be empty?" do
+    #         @that.must_be :empty?
+    #       end
+    #
+    #       it "must have size equal to 0" do
+    #         @that.size.must_equal 0
+    #       end
+    #     end
+    #   end
+    #
+    #   describe "array" do
+    #     def double; @this + @this end
+    #
+    #     before do
+    #       @this = []
+    #     end
+    #
+    #     include EmptyBehavior
+    #   end
+    #
+    #   describe "hash" do
+    #     def double; @this.merge(@this) end
+    #
+    #     before do
+    #       @this = {}
+    #     end
+    #
+    #     include EmptyBehavior
+    #   end
+
+    def shared_description *desc, &block
+      raise TypeError, "can only call shared_description in a module, not a class" if is_a?(Class)
+      shared_descriptions << [desc, block]
+    end
+
+    ##
+    # For each of the modules given, if the included module has any shared descriptions
+    # and this is a class, call describe with all of the shared descriptions in the module.
+    # If the included module has any shared descriptions and this is a module,
+    # include the include module's shared descriptions in the current module's shared
+    # descriptions, so that classes that include the current module get the shared descriptions
+    # in the included modules.
+
+    def include(*mods)
+      res = super
+      mods.each do |mod|
+        if mod.respond_to?(:shared_descriptions)
+          if is_a?(Class)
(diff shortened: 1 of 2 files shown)


zenspider:
I suspect I don't understand your goal, or at the least, your style of testing/thinking-about-testing.

> Currently, if you want to share spec subclasses across multiple specs, you end up having to do something ugly like: [code cut]

No, you don't. You make a module with the tests you want and you include it where you want it. Plain ruby. Remove the `shared_description` block from your example above, and it works fine.

I honestly can't make heads nor tails out of this: https://github.com/jeremyevans/minitest/blob/shared_description/test/minitest/test_minitest_spec.rb#L827-L852

I would try to make a counter-argument in code, but I can't w/o understanding the whats and whys of that code.

jeremyevans:
No, removing the `shared_description` block does not work fine, it breaks the tests (did you actually try running them?).  Without the `shared_description` block, the before/after hooks in the included module run outside the before/after hooks in the class, instead of inside.

Let's say you want to test some behavior for a database, and you test it:

1) Outside of a transaction
2) Inside of a transaction
3) Inside of a savepoint inside a transaction

Without shared descriptions, you basically have to use `self.included` hacks as I pointed out in the commit message.  Apologies for the longish example, but with shared descriptions you can do:

```ruby
module SomeBehavior
  extend Minitest::Spec::DSL
  # lots of specs
end

module InsideSavepoint
  extend Minitest::Spec::DSL
  shared_description "inside savepoint" do
    before do
      # Start Savepoint
    end
    after do
       # End Savepoint
    end
    include SomeBehavior
  end
end

module TransactionBehavior
  extend Minitest::Spec::DSL
  before do
    # Start Transaction
  end
  after do
     # End Transaction
  end
end

module InsideTransaction
  extend Minitest::Spec::DSL
  shared_description "inside transaction" do
    include TransactionBehavior
    include SomeBehavior
  end
  shared_description "inside transaction" do
    include TransactionBehavior
    include InsideSavepoint
  end
end

describe "some behavior" do
  before do
    # some code to be executed before transaction
  end
  after do
    # some code to be executed after transaction
  end
  include InsideTransaction
  include SomeBehavior
end

describe "some other behavior" do
  before do
    # different code to be executed before transaction
  end
  after do
    # different code to be executed after transaction
  end
  include InsideTransaction
  include SomeBehavior
end
```

Due to how ruby method lookup works, before/after blocks in included modules in spec classes occurs outside the before/after blocks in the spec class itself.  There are cases such as the one above where that cannot work, you need before/after in the included modules to run inside the the before/after blocks of the spec class.  You can do this by subclassing, and shared descriptions allow you to share code that defines spec subclasses.

jeremyevans:
Since this is simple to implement as a external extension to minitest, and I thought of a better API, I'm going to close this issue now.  I'll submit a separate pull request to add the external extension to the README.

PR #548, opened 2015-05-18
add a no-op xit

presidentJFK:


teoljungberg:
This could easily be extracted into a plugin. I would suggest using `skip` instead of printing to stdout

PR #550, opened 2015-05-31
Option to not shuffle suites

pjump:
RATIONALE: http://stackoverflow.com/questions/30558791/minitest-i-suck-and-my-tests-are-order-dependent

USAGE:
`Minitest.should_shuffle_suites = false` to disable shuffling of suites.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -13,8 +13,11 @@ module Minitest
   @@installed_at_exit ||= false
   @@after_run = []
   @extensions = []
+  @should_shuffle_suites = true
 
   mc = (class << self; self; end)
+
+  mc.send :attr_accessor, :should_shuffle_suites
 
   ##
   # Parallel test executor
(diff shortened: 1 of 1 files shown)


zenspider:
I have severe reservations about this...

pjump:
I like how minitest shuffles test. It helps me ensure I don't end up building up state in my test suite and depending on it.  But I think it's something that the user should be able to turn off, completely, not just partially. Turning off test-method shuffling is already possible. It'd only be consistent to add the option to turn off suite shuffling too. As I say in the SO post, it'd be mainly good for tests that double as documentation (because randomized documentation kind of sucks). IMO, tools should give the power to users.

zenspider:
I'm leaning severely towards "no". But know that I'm still thinking about this.

zenspider:
Nope. I just can't. Sorry.

First, I'm very much NOT in the camp of "tests generate documentation". I have _never ever_ seen an example that was 1) valid 2) good or 3) *actually* used. Second, the second you put that assignment in your suite and fix your tests to ordered, it introduces the ability to have test order dependency bugs. Minitest was the first test framework to randomize tests as the default, and it was a bug/mistake to not do the suites at the same time. I don't want to introduce something that allows for regressions. Finally, this affects the test runs, NOT the report generation. There can be some other sort of tooling if you want to generate documentation from your tests (but, see first point).

Sorry. I think your argument was well stated, I just wholly disagree with the consequences.

blowmage:
I really like the idea of moving the ordered display of test output to a reporter.

grosser:
FYI This is supported via `Maxitest.static_class_order = true` in [maxitext](https://rubygems.org/gems/maxitest)

PR #552, opened 2015-06-05
Return early if there are no filtered methods

jeremyevans:
This avoids setting up INFO signal handlers for every spec class,
even if there are no methods that will be run inside it.

Additionally, for runnables that override with_info_handler, to
record the number of suites or run code before/after/around suites,
this avoids running other unnecessary code.


zenspider:
Not sure how you come up with this stuff... but OK. :P

zenspider:
merged. Thanks!

PR #561, opened 2015-07-07
Don't extend io with pride if io is not a tty

toy:
Otherwise running a test from TextMate or something else not understanding ANSI codes makes output much less readable.


zenspider:
Love it. Thanks!

PR #563, opened 2015-07-12
fix markup for README.rdoc

yokoto:
This extension of  README is rdoc, but I found markups for md and those were not converted. I fixed wrong markup. If you prefer md, let me know and I will modify it.


zenspider:
I do prefer md, but rdoc's markdown support is SOOO SLOW that I just can't.

zenspider:
Done. Thanks!

Issue #562, opened 2015-07-07
Should assert_raises really rescue Exception ?


francoisjacques:
The current implementation of assert_raises catches everything, which is reasonable for such test, but it is also likely to catch Interrupt which could have been raised during the block execution, and interferes with normal interrupt handling, if Ctrl-C is hit during the block execution.

I seem to get better results by slightly tweaking it with:

```
   rescue Minitest::Skip => e
        return e if exp.include? Minitest::Skip
        raise e
+   rescue SystemExit, Interrupt => e
+     raise e
     rescue Exception => e
```

Thoughts?

zenspider:
> "it is also likely to catch Interrupt which could have been raised during the block execution"

It is also possible to have been raised by the code on purpose:

```
ruby -e 'begin; exit(1); rescue Exception => e; p e; end'
#<SystemExit: exit>
```

I would be hesitant to make your change without some extra logic to make it possible to still test for those exceptions.

francoisjacques:
Ahh yes, true. Darn. Let's think about it some more. There's always the possibility to make it an opt-in option... but it would make the assert_raises syntax clumsy.

zenspider:
Please close this if you think it has been resolved.

francoisjacques:
It has been resolved indeed! Thanks!

PR #565, opened 2015-07-17
Assert raises refactor

bhenderson:
the two commits are kind of different, if you like one and not the other. This is mostly for thinking about #562. I don't really have any stake in the matter.


zenspider:
I'm not entirely sure about the static list o' exceptions... probably should use PASSTHROUGH_EXCEPTIONS or somesuch. That said, I've merged this. Thanks!

PR #566, opened 2015-07-18
Bug in Mock#verify

pithub:
I noticed a bug in `Mock#verify`:

```ruby
mock = Minitest::Mock.new
mock.expect :foo, nil, [:bar]
mock.expect :foo, nil, [:bar]
mock.foo :bar
mock.verify
```

doesn't raise a `MockExpectationError`. Here's a fix consisting of the following commits:

* **add failing test for bug** - the test specifies an error message which is not optimal, but matches the current implementation (see below)
* **fix bug**
* **specify better error messages in tests** - improved the expected error messages (at least in my eyes) - _this has an effect on an existing test!_
* **fix tests** - since the history of actual calls isn't needed anymore for the new error messages, some code could be removed
* **fix method rdoc of expect** - the documentation didn't describe the current behavior

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -65,8 +65,7 @@ module Minitest # :nodoc:
     #   @mock.verify  # => true
     #
     #   @mock.expect(:uses_one_string, true, ["foo"])
-    #   @mock.uses_one_string("bar") # => true
-    #   @mock.verify  # => raises MockExpectationError
+    #   @mock.uses_one_string("bar") # => raises MockExpectationError
 
     def expect(name, retval, args = [], &blk)
       name = name.to_sym
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -96,16 +96,11 @@ module Minitest # :nodoc:
     # expected.
 
     def verify
-      @expected_calls.each do |name, calls|
-        calls.each do |expected|
-          raise MockExpectationError, "expected #{__call name, expected}, got [#{__call name, @actual_calls[name]}]" if
-            @actual_calls.key?(name) and
-            not @actual_calls[name].include?(expected)
-
-          raise MockExpectationError, "expected #{__call name, expected}" unless
-            @actual_calls.key?(name) and
-            @actual_calls[name].include?(expected)
-        end
+      @expected_calls.each do |name, expected|
+        actual = @actual_calls.fetch(name, nil)
+        raise MockExpectationError, "expected #{__call name, expected[0]}" unless actual
+        raise MockExpectationError, "expected #{__call name, expected[actual.size]}, got [#{__call name, actual}]" if
+          actual.size < expected.size
       end
       true
     end
--- a/test/minitest/test_minitest_mock.rb
+++ b/test/minitest/test_minitest_mock.rb
@@ -240,6 +240,20 @@ class TestMinitestMock < Minitest::Test
     assert_equal exp, e.message
   end
 
+  def test_same_method_expects_with_same_args_blow_up_when_not_all_called
+    mock = Minitest::Mock.new
+    mock.expect :foo, nil, [:bar]
+    mock.expect :foo, nil, [:bar]
+
+    mock.foo :bar
+
+    e = assert_raises(MockExpectationError) { mock.verify }
+
+    exp = "expected foo(:bar) => nil, got [foo(:bar) => nil]"
+
+    assert_equal exp, e.message
+  end
+
   def test_verify_passes_when_mock_block_returns_true
     mock = Minitest::Mock.new
     mock.expect :foo, nil do


zenspider:
I'm not sure I'd call this a bug or a design choice... but you've removed more code than you added, so I'm a fan. :)

@zamith, opinion?

zamith:
This is definitely a design choice. However, this is the second time (at least) that someone has requested mocks to have a knowledge of how many times a method is called.

So, even though that might be something we want to tackle, I'd say removing the verify step entirely is not a good way of achieving that. With these changes, the `verify` method is superfluous and only there for compliance reasons, also we loose the knowledge we had of the calls that were actually made, thus loosing some ability to provide insightful error messages to the users.

So, even though the underlying problem might be considered a bug, the solution changes the way mocks work a bit too much, in my opinion.

pithub:
Thanks for looking into this.

As to whether this is a bug or a design choice: don't you think that `verify` should be consistent with the code verifying the actual method calls? When you `expect` a method n times, then it is an error to actually call it for the n+1st time. So mocks already have a knowledge of how many times a method is called. If it is an error to call a method too often, I'd expect it to be an error to call it too rarely, too. This would be recognized by `verify`, so `verify` definitely would still have a reason.

As to whether too much code would be removed: of course it would be possible to leave the knowledge of method calls that were actually made, but I didn't see any actual usage of that, so I removed it.

zamith:
Yes, at this point I'd say that is a bug, however I think we'll be better off by keeping the actual calls information, or something similar, that allows us to provide better errors, such as "you expected X to be called, but Y was called", or "you expected X to be called once, but it was called twice".

So, my concerns here are basically that we have enough information so that we can provide good, insightful error messages. If we can do that and not keep track of the actual calls, then by all means let's remove them.

pithub:
@zamith: the error messages you've shown should be possible without having the history of actual calls, because they only need the actual and the next expected call, which are still available. You couldn't provide error messages that really need the history like "you expected X and then Y to be called, but X and then Z was called".

zamith:
How would you know if a call was made 3 or 4 times, while you only expected it to be called once?

pithub:
Sorry, you're too fast for me :-)  I edited my last comment while you responded. Could you look at my last comment again?

zamith:
Yes, I also thought we could maybe do it with a counter. The only other thing I have been thinking about is if we should have an explicit way of saying we expect this call to be made 3 times.

Maybe:

```
mock.expect :foo, nil, [:bar], times: 3
```

Anyway, would you be able to change this PR to provide better error messages, at least in the situations I mentioned? Other than that, I'd be good with it, since it does not change any public facing API, we can always revert to saving the actual calls if we need to.

zamith:
@pithub Any news on this? Are you still interested in this change?

pithub:
@zamith Yes, still interested. Back in September, the more I thought about better error messages, the more I got the impression that this might be a design choice, after all. Then I got sidetracked by day-to-day work... Things might improve next week, though. I'll sort and post the thoughts I had, so we can decide how to proceed.

pithub:
@zamith OK, after the *third* code review I think I *finally* got the intention of the current implementation. I have to say that I didn't really get it when writing my patch. My problem was that I read the error message
  `expected foo(:baz) => nil, got [foo(:bar) => nil]`
as
  `expected to get ..., BUT got ... INSTEAD`
and not as
  `expected to get ..., HAVING got ... SO FAR`

Interpreted this way,
  `expected foo(:bar) => nil, got [foo(:bar) => nil]`
would be a perfectly valid and meaningful error message.

So I'd like to leave the error messages as they are and submit only the first two of my commits - and maybe the last one (correction of a comment). What do you think of this? Do you want me to look at your pull request too?

zamith:
I like this, I think it simplifies the internals, without changing the public facing API.

@zenspider What do you think? Want to merge this in?

zenspider:
@zamith if you (still) like it, that's good enough for me.

Where does this stand? Merge?

zamith:
@zenspider My only concern is that this is a breaking change (I think), because now instead of raising only on the `verify`, we raise on the exercise part of test, whenever the unexpected method is called.

However, I like the approach because it fails faster, which is a good thing for tests, while cleaning up the code base.

So, to sum up, I like it, but not sure if it will break for some users. If you don't think that is an issue, I'd say merge.

zenspider:
> @zenspider My only concern is that this is a breaking change (I think), because now instead of raising only on the verify, we raise on the exercise part of test, whenever the unexpected method is called.

Can you point me at a test case for this? I'm not seeing it (yet--need coffee).

zenspider:
Ignoring my confusion above, my only concern about this patch is this part:

```diff
-    exp = "expected foo(:baz) => nil, got [foo(:bar) => nil]"
+    exp = "expected foo(:baz) => nil"
```

Why not say what we have gotten so far?

zamith:
The thing here is about calling an expected method twice, and how it should error on the second time, which I think it makes sense.

Now that you mention it, I think that it is important to say what we have gotten as well as what we were expecting. @pithub

zamith:
I don't think there is the need for a new pull request. At this point, the only issue is that the error message is different. If you can make it not change, we should be good to go.

lib/minitest/mock.rb
```diff
@@ -97,14 +86,9 @@ def __call name, data # :nodoc:
...
       @expected_calls.each do |name, calls|
-        calls.each do |expected|
-          raise MockExpectationError, "expected #{__call name, expected}, got [#{__call name, @actual_calls[name]}]" if
-            @actual_calls.key?(name) and
-            not @actual_calls[name].include?(expected)
-
-          raise MockExpectationError, "expected #{__call name, expected}" unless
-            @actual_calls.key?(name) and
-            @actual_calls[name].include?(expected)
+        if next_call = calls.first
```
zenspider: mmm... no assignment in condition pls.

zenspider:
It looks like the removal of `actual_calls` prevents the new system from printing out what it has gotten. I can't easily get it to work cleanly.

zenspider:
this is my only current change against the patch:

```diff
diff -c /Users/ryan/Work/p4/zss/src/minitest/dev/lib/minitest/mock.rb\~ /Users/ryan/Work/p4/zss/src/minitest/dev/lib/minitest/mock.rb
--- /Users/ryan/Work/p4/zss/src/minitest/dev/lib/minitest/mock.rb~	2015-11-16 11:09:10.000000000 -0600
+++ /Users/ryan/Work/p4/zss/src/minitest/dev/lib/minitest/mock.rb	2015-11-17 03:05:27.000000000 -0600
@@ -86,7 +86,9 @@
 
     def verify
       @expected_calls.each do |name, calls|
-        if next_call = calls.first
+        next_call = calls.first
+
+        if next_call then
           raise MockExpectationError, "expected %s(%s) => %p" %
             [name, next_call[:args].inspect[1..-2], next_call[:retval]]
         end

```

pithub:
@zenspider @zamith Please read my [last comment] (https://github.com/seattlerb/minitest/pull/566#issuecomment-157162704). Originally the patch consisted of five commits, but in the meantime I withdraw the third and the fourth commit. (That's why I offered to change the pull request.) With the remaining commits, there'd be no change of error messages and no assignment in conditions.

zamith:
@pithub can you just rebase this then, so that it is the right code?

pithub:
@zamith Sure, but you have to wait until this evening (CET).

zamith:
That's fine.

zamith:
@pithub I like it. It's cleaner, and there should be no changes to the user interface, which is nice. @zenspider Do you have any comments? If not, I'd say :shipit:.

zenspider:
Done and released! Thanks!

pithub:
Thank you both for your patience ☺

PR #567, opened 2015-07-23
Change refute default message to include object

bhenderson:
I run accross this when any result signals an error. ie.

> refute flash.now[:alert]
> Failed refutation, no message given # but what was the error?

instead of

> refute flash.now[:alert], flash.now[:alert]
> Error in blah blah... # now I know my problem

I don't think it makes sense for assert because 'Expected nil to be
truthy' is not helpful

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -500,7 +500,7 @@ module Minitest
     # Fails if +test+ is truthy.
 
     def refute test, msg = nil
-      msg ||= "Failed refutation, no message given"
+      msg ||= "Expected #{mu_pp(test)} to not be truthy"
       not assert !test, msg
     end
 
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -500,7 +500,7 @@ module Minitest
     # Fails if +test+ is truthy.
 
     def refute test, msg = nil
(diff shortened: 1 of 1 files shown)


zenspider:
It's a lot less info, but do you think that `assert`, should have a similar change?

bhenderson:
`Expected nil to be truthy`
vs
`Failed assertion, no message given`

```ruby
class Foo
  def something() @something end
end

assert Foo.new.something
# Expected {Foo.new.something} to be truthy
```
something like that could be helpful, but I don't think I want to get in that mess.

lib/minitest/assertions.rb
```diff
@@ -500,7 +500,7 @@ def pass _msg = nil
     # Fails if +test+ is truthy.
 
     def refute test, msg = nil
-      msg ||= "Failed refutation, no message given"
+      msg ||= "Expected #{mu_pp(test)} to not be truthy"
```
zenspider: hrm. as soon as you add interpolation, you should wrap it in a `message { ... }` block.

zenspider:
poking... I think I still want to add something similar to assert.

zenspider:
OK. Done. Thanks! Let me know what you think of my mods.

bhenderson:
looks great. thanks

PR #571, opened 2015-08-09
add assert_not_equal method (and unit test for it)

sebhtml:


zenspider:
I can not accept this PR. Please go watch any of my minitest talks for a thorough explanation why.

sebhtml:
Can you provide a hyper link to one of these talks ?

sebhtml:
I see. This is neat. I have to use refute_equal.

http://www.zenspider.com/pdf/2015_railsconf_ruby_on_rails_on_minitest.pdf

Thanks for pointing this.

zenspider:
_By itself_, `refute_equal` is a bad assertion. Only use it to contrast a difference with a previous `assert_equal`.

Issue #519, opened 2015-04-13
Add negative filtering so tools like minitest-bisect can work better


zenspider:
Sometimes it'd be better to exclude what we don't want to run. It'd allow all tests to be considered during randomization, but then filtering out the tests we don't want. This allows stable order runs while minitest-bisect filters out entire files.

Negative filtering should be able to do both tests by name (opposite of `--name /regexp/`) as well as entire files (which will require being able to map files to tests?).

zenspider:
Assigning to myself so it stays on my radar, but **please** feel free to submit a pull request for this.

utilum:
Do you see test name and file filtering triggered by a single option?

zenspider:
Probably not. That'd probably be confusing. `--exclude` and `--exclude-file` maybe?

utilum:
@zenspider 
- Is it cool to create an actual file to test file filtering?
- I filter by file name, but perhaps duplicates may exist and it'd be better to use a longer path?

zenspider:
> On Sep 15, 2015, at 22:55, utilum <notifications@github.com> wrote:
> 
> @zenspider
> 
> 	• Is it cool to create an actual file to test file filtering?
> 	• I filter by file name, but perhaps duplicates may exist and it'd be better to use a longer path?

Absolutely cool to create an actual file to test file filtering. Maybe under test/fixtures or test/data or somesuch?

Mmmmmm... don't worry about duplicates for now.

utilum:
Um... actually, there's a compatibility issue in this last PR. Better coming up.

zenspider:
Closing. This is basically done, just not merged because of the issues in #609.

PR #579, opened 2015-09-22
Add negative filters

utilum:


utilum:
#519

zenspider:
OK. I need to ponder. This requires dropping compatibility. I'd like to get numbers to make an informed decision.

utilum:
Dropping compatibility with 1.8.7 sounds harsh. What if I move the file filtering into a plugin that could be combined with tools like minitest-bisect when needed ?

zenspider:
I'm using this right now with just `-e` and I love it. We could forego the file exclusion thing for a plugin for now?

utilum:
Yes, that's the idea. See pull request 602.

PR #580, opened 2015-10-01
Aligned benchmark output in README.rdoc

artlugovoy:


zenspider:
The output from minitest/benchmark is tab separated so you can copy/paste it into a spreadsheet. It isn't aligned per se.

zamith:
That tab thing is a nice trick. Didn't know about it. :+1:

artlugovoy:
Ah, okay. So it would be better just to close it for this reason

PR #581, opened 2015-10-02
js - make it possible to dynamically adjust the order of tests within a run

spjsschl:
This PR makes it possible to make a scheduling decision about which test to run next at runtime.

At my company we run a lot of high level tests (particularly selenium tests) using Minitest as the underlying framework. Tests at such a high level tend to flake a lot more then simple unit tests. Transient network errors, some JS being a little slow, a browser having a cold cache and many more things outside of the programmers control can happen.
We talked to other companies and this is a common problem. Even after addressing as many sources of flakiness as possible and making the individual tests very resilient (extra waiting, retrying, etc) everyone seems to still have the occasional flake. Multiple people across companies have gravitated to the solution of just rerunning a test that failed up to twice more. If the test passes a subsequent run it does not count as failed. We have had our own patches to Minitest for years accomplishing this and I recently rewrote all of that when upgrading to Minitest 5. We thought (since it is a common problem) it would be great if Minitest had some support for making scheduling decisions for tests at runtime. This scheduler can be combined with a reporter to rerun a failed test and report it as passing or failing depending on the result of this rerun.

Currently supported scheduling strategies like :random or :parallel can be expressed in these terms too, but would require some more changes than what I did in this PR. I tried to stay backwards compatible and minimally invasive. My solution is such only a suggestion or starting point for discussion if you will.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -292,7 +292,7 @@ module Minitest
       return if filtered_methods.empty?
 
       with_info_handler reporter do
-        filtered_methods.each do |method_name|
+        method_stream(filtered_methods) do |method_name|
           run_one_method self, method_name, reporter
         end
       end
@@ -306,6 +306,10 @@ module Minitest
 
     def self.run_one_method klass, method_name, reporter
       reporter.record Minitest.run_one_method(klass, method_name)
+    end
+
+    def self.method_stream(methods)
+      methods.each {|m| yield m}
     end
 
     def self.with_info_handler reporter, &block # :nodoc:
--- a/test/minitest/test_minitest_unit.rb
+++ b/test/minitest/test_minitest_unit.rb
@@ -284,6 +284,78 @@ class TestMinitestRunner < MetaMetaMetaTestCase
     assert_report expected, %w[--name /some|thing/ --seed 42]
   end
 
+  def test_run_default_order
+    @tu =
+        Class.new Minitest::Test do
+          @order = []
+
+          def self.order
+            instance_variable_get(:@order)
+          end
+
+          def test_some_other_thing
+            self.class.order << :test_some_other_thing
+            assert_equal [:test_some_other_thing], self.class.order
+          end
+
+          def test_something
+            self.class.order << :test_something
+            assert_equal [:test_some_other_thing, :test_something], self.class.order
+          end
+        end
+
+    expected = clean <<-EOM
+      ..
+
+      Finished in 0.00
+
+      2 runs, 2 assertions, 0 failures, 0 errors, 0 skips
+    EOM
+
+    assert_report expected, %w[--seed 42]
+  end
+
+  def test_run_dynamic_order
+    @tu =
+        Class.new Minitest::Test do
+          def self.method_stream(_)
+            ([:test_passes_eventually] * 2).each do |method|
+              yield method
+            end
+          end
+
+          @count = 0
+
+          def self.count
+            instance_variable_get(:@count)
+          end
+
+          def self.count=(c)
+            instance_variable_set(:@count, c)
+          end
+
+          def test_passes_eventually
+            self.class.count = self.class.count + 1
+            assert_equal 2, self.class.count
+          end
+        end
+
+    expected = clean <<-EOM
+      F.
+
+      Finished in 0.00
+
+        1) Failure:
+      #<Class:0xXXX>#test_passes_eventually [FILE:LINE]:
+      Expected: 2
+        Actual: 1
+
+      2 runs, 2 assertions, 1 failures, 0 errors, 0 skips
+    EOM
+
+    assert_report expected, %w[--seed 42]
+  end
+
   def assert_filtering name, expected, a = false
     args = %W[--name #{name} --seed 42]
 


zenspider:
I'm not against this. Still weighing the pros and cons... but I was curious why you're not overriding run_one_method instead? Something like:

```ruby
    def self.run_one_method klass, method_name, reporter
      result = nil
      3.times do
        result = Minitest.run_one_method(klass, method_name)
        break if result.passed?
      end
      reporter.record result
    end
```

This records ONLY one entry regardless of how many are run. If you don't mind multiple test runs getting reported (and by your patch, you don't, then you can simplify using super:

```ruby
    def self.run_one_method klass, method_name, reporter
      3.times do
        return if super.passed?
      end
    end
```

This last one requires a bit of refactoring so that `Reporter#record` returns the test object back.

spjsschl:
Hi @zenspider, thank you for the quick reply.
The reason why my PR suggests a different solution, was that I wanted to keep it minimal. I was under the assumption that one would couple this with a custom reporter that would not count/report on tests that were rerun. This way they are not reported multiple times.

My original implementation, that I use at work does patch run_one_method like you suggested, but it is more intrusive:

```
module MiniTest
  class FlakeyReRunner
    def initialize
      self.total_failures = 0
      self.max_failures = 15
      self.max_retries = 3
    end

    def run_one_method(klass, method_name, reporter)
      if total_failures < max_failures && !is_functional?(klass)
        result = run_with_retry(klass, method_name)
      else
        result = run_once(klass, method_name)
      end
      reporter.record(result)
    end

    private

    attr_accessor :total_failures, :max_failures, :max_retries

    def run_with_retry(klass, method_name)
      first_result = nil
      max_retries.times do
        result = run_once(klass, method_name)
        return result if result.passed? || retries_off?(result)
        first_result ||= result
        self.total_failures = total_failures + 1
      end
      first_result
    end

    def run_once(klass, method_name)
      Minitest.run_one_method(klass, method_name)
    end

    def is_functional?(klass)
      klass.ancestors.map(&:to_s).include?('ActionController::TestCase')
    end

    def retries_off?(runnable)
      runnable.instance_variable_get(:@do_not_recover_from_flakiness)
    end
  end

  def self.flakey_re_runner_instance
    @flakey_runner_instance ||= FlakeyReRunner.new
  end

  class Runnable
    def self.run_one_method(klass, method_name, reporter)
      AfTesting::TeamCity::MiniTest.flakey_re_runner_instance.run_one_method(klass, method_name, reporter)
    end
  end

  class Test < Runnable
    def do_not_recover_from_flakiness!
      @do_not_recover_from_flakiness = true
    end
  end
end
```

This solution has a couple of other features:
- it can be selectively disabled
- it reports always the first failure if a test fails all its tries (this is to not report errors inadvertently that were caused because the test can not be rerun)
- it does not rerun Rails functional tests since they do not support it
- it does have a limit on how many tests to rerun in case a suite has some non-recoverable error and would otherwise run up to 3 times as long (we have seen this happen with database crashes for example)

zenspider:
That's... so long. :P

Yes, you're keeping your patch minimal... but no code is even more minimal. `Runnable.run_one_method` is factored that way for a reason. Your original approach suffers from being way too long/complex and "solves" a bunch of "problems" by adding features you don't really need. Compare:

```ruby
require "minitest/autorun"

module Retry
  def run_one_method klass, method_name, reporter
    3.times do
      result = Minitest.run_one_method(klass, method_name)
      reporter.record result and break if result.passed?
    end
  end
end

class TestFlakyStuff < Minitest::Test
  extend Retry

  def setup
    @@n ||= 0
  end

  def test_something_bad
    @@n += 1
    assert(@@n >= 3)
  end
end
```

It's opt-in, not opt-out, so if you don't want it on in a particular class, don't add it. Functional tests don't like retries? Don't extend them with Retry.

Still, I'm considering your enumerator approach. I'm not against it, but I'm not sure it is needed or provides tangible benefit the more I think about it.

Finally, this really feels like a bandaid applied to cover up a symptom, not actually solving the real problem. Tests must be consistent and reproducible in order to be trustworthy. If you need something like this, something's wrong. If your tests aren't stable, what are you going to do in production?

spjsschl:
I like your solution, that is pretty much what I started with. I ended up with feature creep to this solution for company internal reasons. I am aware of that and that is why I did not make this my PR.

I was thinking the enumerator approach enables arbitrary decisions about test order at runtime, maybe there are others out there that have need for this but you know that probably better than I do.

I totally agree with you that the easiest solution for the rerunner is patching run_one_method.

For your last point, point taken and in a perfect world I very much agree with you. Yet business decisions are usually tradeoffs. A lot of these problems causing flakiness are hard to debug. While we try to solve them, at the end of the day if one of them happens in production we have someone follow up and fix the real world damage done if any was done. It is just cheaper than having well paid engineers figuring these out at a rate of one a week. But again, I hear you.

pkmiec:
@spjsschl @zenspider Couple weeks ago, I was at the SeleniumConf 2015. It seems everyone is struggling with flakiness in selenium tests. The common / recommended ways to minimize flakiness is by having solid layer to interface with selenium (e.g. capybara in the ruby world), solid page objects and helpers, and test patterns that make it easier for test writers to avoid pitfalls. But this only minimizes flakiness to a certain point. The next recommended method is to re-try failed test a number of times and only really fail the test if it fails 3 times.

Since it's a common problem, it would be great if minitest had some support for it. Perhaps there could be an interface or stable extension point(s) in minitest to support it (ala reporters). And then we could have a minitest-auto_retry extension which could be enabled for certain suites (e.g. `Minitest::AutoRetry.configure do |config| ... end`).

Maybe another use of such an extension point would be to support a different parallel strategy. One where tests are parallelized between processes or between difference hosts.

I don't know the internals of minitest, so I can't get into specifics.

zenspider:
@pkmiec My point is that support _is in minitest_ for this. The design explicitly allows you to change how a test is run. The code above (and the new adaptation below) makes it beyond easy to do such a thing.

```ruby
def Retry n
  Module.new do
    define_method :run_one_method do |klass, method_name, reporter|
      result = nil
      n.times do
        result = Minitest.run_one_method(klass, method_name)
        break if result.passed?
      end
      reporter.record result
    end
  end
end

class TestFlakyStuff < Minitest::Test
  extend Retry(3)

  # ... unstable tests ...
end
```

zenspider:
@spjsschl wrt your last paragraph... please don't read my final point like I'm trying to put this onus or its solution on you. I'm not. I know that things like selenium are untrustworthy. We just need to make sure that we're not setting up a situation where by enabling rerunning unstable tests, we wind up putting bad code into production. Put too many bandaids on, and you won't know where things went wrong because you smoothed them over too much.

pkmiec:
@zenspider i see .. great.

trabulmonkee:
@zenspider I like the latest version of the Retry method above.  One thing to consider are skips to break on rather than running the skip multiple times.  Something like?
```ruby
def Retry n
  Module.new do
    define_method :run_one_method do |klass, method_name, reporter|
      result = nil
      n.times do
        result = Minitest.run_one_method(klass, method_name)
        break if result.passed? or result.skipped?
      end
      reporter.record result
    end
  end
end

class TestFlakyStuff < Minitest::Test
  extend Retry(3)

  # ... unstable tests ...
end

spjsschl:
Sorry for being inactive here a little while. I will go ahead and make a plugin that encapsulates the retry logic as discussed above. I will leave a link to the repo here once I have it.
Maybe that can go into the extensions-list then @zenspider?

spjsschl:
I found above proposed and discussed solution pretty much already exists: https://github.com/y-yagi/minitest-retry

Maybe this is worthy putting on the plugin list.

blowmage:
I work on a project that runs acceptance tests against live services, and those services can be flakey. We used to restart builds several times a day. But using @zenspider's approach has made failing builds the exception rather than the rule. Can't recommend it enough.

What I like about the approach is that we can decide which tests to retry. The minitest-retry gem seems to be an all or nothing approach. All tests are retried or none of them are. And it seems like it may trip over itself when running tests in parallel, not sure though. For our tests we simply override `run_one_method` on the test classes we want to retry failures on. Works great. :+1:

https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/acceptance/datastore_helper.rb#L39-L47

spjsschl:
thanks for the feedback @blowmage. Sounds like there is interest in a second solution. I have a git repo ready, I just need to rename it before releasing the first version of the gem, to avoid the name conflict with y-yagi's gem

spjsschl:
I am done with renaming and have released the first version to rubygems. It would be great if this gem makes it on the plugin list. From my perspective the PR can be closed.

https://rubygems.org/gems/minitest-optional_retry
https://github.com/appfolio/minitest-optional_retry

zenspider:
Fantastic. Feel free to send me a PR adding your gem to the list!

zenspider:
This is closable, yes? Please close if you're satisfied or comment if something is still up in the air.

spjsschl:
yes

thanks,
        Johannes

PR #586, opened 2015-10-09
Assert match only counts as one assertion

### LINKED ISSUE #584: assert_match should only increment assertions by 1 (not 2)
I just noticed that when I use assert_match, it increments assertions by 2.  I looked in the code to see why, and found the method definition here:

https://github.com/seattlerb/minitest/blob/master/lib/minitest/assertions.rb#L223

    def assert_match matcher, obj, msg = nil
      msg = message(msg) { "Expected #{mu_pp matcher} to match #{mu_pp obj}" }
      assert_respond_to matcher, :"=~"
      matcher = Regexp.new Regexp.escape matcher if String === matcher
      assert matcher =~ obj, msg
    end

I'm new to minitest, so I may have some missing context, but wouldn't it better to use something *other* than an assert* method to ensure the matcher passed in responds to `=~`?  The point of that code is to validate inputs to the assertion, whereas the point of an assertion is to test the behavior of the app code the test suite is being applied to.

Can't this line be replaced with some kind of error checking that isn't based on calling an assert* method, thereby incrementing the counter and confusing the dev writing tests?

zamith:
Why:

* The method `assert_match` was counting as two assertions, for some
reason I could not figure out from the git log.

* Removing one of the assertions
* Still raising an asserting error with the expected message when the
matcher does not respond to the `=~` method, even though there is no
test for that, and I don't really understand why that would be needed.
A method like `assert_output` cleary states in the documentation it will
make two assertions. The `assert_match` method does not, on the
contrary, it claims to make one.

Fixes #584.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -222,7 +222,7 @@ module Minitest
 
     def assert_match matcher, obj, msg = nil
       msg = message(msg) { "Expected #{mu_pp matcher} to match #{mu_pp obj}" }
-      assert_respond_to matcher, :"=~"
+      raise Minitest::Assertion, "Expected #{mu_pp(matcher)} (#{matcher.class}) to respond to =~" unless matcher.respond_to? :"=~"
       matcher = Regexp.new Regexp.escape matcher if String === matcher
       assert matcher =~ obj, msg
     end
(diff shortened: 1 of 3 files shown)


zenspider:
It counts as 2 because it actually checks 2. I'm rejecting #584, so I'm rejecting this. Sorry. I really do appreciate your efforts here.

zamith:
No problem, I found it a bit odd at first, but I get the "debugability" issue, and in reality no one looks at the number of assertions anyway, I'd say.

Well, I found about `assert_output`, so not all is lost. ;)

PR #588, opened 2015-10-12
Adds option to verify the number of times a call is made

zamith:
Why:

* It seems useful to be able to do this, especially in the situation
where you want to verify a method is never called, because that wasn't
even possible

* Storing the expected times with the expected call
* Making sure method_missing allows the call to go through when we
expect it more than once, even though we have only one record on the
expected_calls array, while also making sure it will fail if we aren't
expecting it more than once
* Doing most of the work on verify, to provide a nice error when the
number of times we expect a method to be called is not met. This is
opt-in, if you do not add a times option, the verify method won't even
touch this new code

As discussed in #566.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -68,15 +68,15 @@ module Minitest # :nodoc:
     #   @mock.uses_one_string("bar") # => true
     #   @mock.verify  # => raises MockExpectationError
 
-    def expect(name, retval, args = [], times: nil, &blk)
+    def expect(name, retval, args = [], opts = {}, &blk)
       name = name.to_sym
 
       if block_given?
         raise ArgumentError, "args ignored when block given" unless args.empty?
-        @expected_calls[name] << { :retval => retval, :times => times, :block => blk }
+        @expected_calls[name] << { :retval => retval, :times => opts[:times], :block => blk }
       else
         raise ArgumentError, "args must be an array" unless Array === args
-        @expected_calls[name] << { :retval => retval, :times => times, :args => args }
+        @expected_calls[name] << { :retval => retval, :times => opts[:times], :args => args }
       end
       self
     end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -68,15 +68,15 @@ module Minitest # :nodoc:
     #   @mock.uses_one_string("bar") # => true
     #   @mock.verify  # => raises MockExpectationError
 
-    def expect(name, retval, args = [], &blk)
+    def expect(name, retval, args = [], times: nil, &blk)
       name = name.to_sym
 
       if block_given?
         raise ArgumentError, "args ignored when block given" unless args.empty?
-        @expected_calls[name] << { :retval => retval, :block => blk }
+        @expected_calls[name] << { :retval => retval, :times => times, :block => blk }
       else
         raise ArgumentError, "args must be an array" unless Array === args
-        @expected_calls[name] << { :retval => retval, :args => args }
+        @expected_calls[name] << { :retval => retval, :times => times, :args => args }
       end
       self
     end
@@ -98,6 +98,14 @@ module Minitest # :nodoc:
     def verify
       @expected_calls.each do |name, calls|
         calls.each do |expected|
+          if expected[:times]
+            number_of_times_called = (@actual_calls.key?(name) && @actual_calls[name].count{ |call| call == expected }) || 0
+            raise MockExpectationError, "expected #{__call name, expected} to be called #{expected[:times]} times but was called #{number_of_times_called} times" unless
+              number_of_times_called == expected[:times]
+
+            next
+          end
+
           raise MockExpectationError, "expected #{__call name, expected}, got [#{__call name, @actual_calls[name]}]" if
             @actual_calls.key?(name) and
             not @actual_calls[name].include?(expected)
@@ -121,15 +129,16 @@ module Minitest # :nodoc:
       end
 
       index = @actual_calls[sym].length
-      expected_call = @expected_calls[sym][index]
+      number_of_expects = @expected_calls[sym].map { |call| call[:times] || 1 }.inject(&:+)
+      expected_call = @expected_calls[sym][index] || @expected_calls[sym].last
 
-      unless expected_call then
+      unless expected_call && index < number_of_expects then
         raise MockExpectationError, "No more expects available for %p: %p" %
           [sym, args]
       end
 
-      expected_args, retval, val_block =
-        expected_call.values_at(:args, :retval, :block)
+      expected_args, retval, val_block, expected_times =
+        expected_call.values_at(:args, :retval, :block, :times)
 
       if val_block then
         # keep "verify" happy
@@ -159,6 +168,7 @@ module Minitest # :nodoc:
       @actual_calls[sym] << {
         :retval => retval,
         :args => zipped_args.map! { |mod, a| mod === a ? mod : a },
+        :times => expected_times,
       }
 
       retval
--- a/test/minitest/test_minitest_mock.rb
+++ b/test/minitest/test_minitest_mock.rb
@@ -227,17 +227,13 @@ class TestMinitestMock < Minitest::Test
   end
 
   def test_same_method_expects_blow_up_when_not_all_called
-    mock = Minitest::Mock.new
-    mock.expect :foo, nil, [:bar]
-    mock.expect :foo, nil, [:baz]
+    @mock = Minitest::Mock.new
+    @mock.expect :foo, nil, [:bar]
+    @mock.expect :foo, nil, [:baz]
 
-    mock.foo :bar
+    @mock.foo :bar
 
-    e = assert_raises(MockExpectationError) { mock.verify }
-
-    exp = "expected foo(:baz) => nil, got [foo(:bar) => nil]"
-
-    assert_equal exp, e.message
+    util_verify_bad "expected foo(:baz) => nil, got [foo(:bar) => nil]"
   end
 
   def test_verify_passes_when_mock_block_returns_true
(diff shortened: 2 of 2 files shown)


lib/minitest/mock.rb
```diff
@@ -68,15 +68,15 @@ def initialize(delegator = nil) # :nodoc:
     #   @mock.uses_one_string("bar") # => true
     #   @mock.verify  # => raises MockExpectationError
 
-    def expect(name, retval, args = [], &blk)
+    def expect(name, retval, args = [], times: nil, &blk)
```
teoljungberg: Since minitest supports ruby 1.8.7 -> 2.2.* - we can't use keyword arguments. Thoughts on turning this into an options hash?

zamith: Right. Remembered that on the hashes, totally forgot about it here. Thanks for pointing it out.

I think the most important is the interface for the user, so I'd definitely go with an options hash.

test/minitest/test_minitest_mock.rb
```diff
@@ -227,17 +227,13 @@ def test_same_method_expects_are_verified_when_all_called
...
-    mock = Minitest::Mock.new
-    mock.expect :foo, nil, [:bar]
-    mock.expect :foo, nil, [:baz]
-
-    mock.foo :bar
-
-    e = assert_raises(MockExpectationError) { mock.verify }
+    @mock = Minitest::Mock.new
+    @mock.expect :foo, nil, [:bar]
+    @mock.expect :foo, nil, [:baz]
```
zenspider: this chunk of diff (and more below) are needless diffs. Don't use ivars when you don't need to.

test/minitest/test_minitest_mock.rb
```diff
@@ -343,6 +339,30 @@ def test_mock_called_via_send_with_args
...
+    mock.expect(:foo, true, [], times: 0)
+
+    mock.verify
+  end
+
+  def test_mock_blow_up_when_called_less_times
+    @mock = Minitest::Mock.new
+    @mock.expect(:foo, true, [], times: 2)
+
+    @mock.foo
```
zenspider: ditto

zenspider:
> It seems useful to be able to do this, especially in the situation where you want to verify a method is never called, because that wasn't even possible

I'm going to trust your opinion on this. I completely disagree and find this form of testing to not only be limiting, but incredibly brittle (add/remove caching=break tests! woo!). As far as I'm concerned, you own decisions on minitest/mock as you've shown the most interest and match my style for the most part (*, **).

That said, I have issues with the implementation here. While I don't see value in this sort of testing (but I will ignore that for a while) I do take exception with the extra complexity added for the sake of being more "DSL-like", or... whatever. I'm low on coffee and words are hard. I don't find the `times: 2` argument to add that much to readability. The user could just as well call expect twice or use `n.times { ... }`. _[... getting back to this... now I'm post workout and even dumber...]_ Adding support for a looping construct doesn't feel "right".

Maybe when I have more brain I can be more eloquent.

\* We should discuss dropping 1.8 (and 1.9) in a separate ticket.
\*\* We should also discuss removing minitest/mock and maybe even minitest/spec in a separate ticket

test/minitest/test_minitest_mock.rb
zenspider: looking at the whole file now, this change makes more sense. I'm starting to wonder if the setup method shouldn't be unfactored as it really isn't saving much of anything. I just have a kneejerk reaction to `@`-only changes.

zenspider:
As it stands, if I merge in the tests and not the impl the only test that fails is:

```ruby
  def test_mock_blow_up_when_called_fewer_times
    @mock = Minitest::Mock.new

    @mock.expect :foo, true, []
    @mock.expect :foo, true, []

    @mock.foo

    util_verify_bad "expected foo() => true to be called 2 times but was called 1 times"
  end
```

(modified to not use `times`).

zamith:
How does this one pass? (also removed the `times`) To be honest, this is the only use case I really need and cannot currently achieve, and will not be able to do with a loop, I think.

```ruby
  def test_mock_was_never_called
    mock = Minitest::Mock.new
    mock.expect(:foo, true, [])

    mock.verify
  end
```

test/minitest/test_minitest_mock.rb
zamith: I saw this and the util method and felt like this was just forgotten, but then again, this does add an instance variable for no good reason, and worse than that will cause git churn for stylistic reasons. So I'll probably just leave it as is. Are you fine with that?

zamith:
Regarding the complexity, I agree, it ended up being more complex than what I was aiming at. I can try to improve it somehow, but I don't think it will get much better with this approach. Maybe incorporating the changes in #566 it can be a bit easier to reason about, because we loose the `@actual_calls` and are just working with one hash, instead of two.

But as I said on the previous comment, the new feature is to test that something is never called, we might narrow it down to just that. What are your thought on that?

zenspider:
Never called already works, no? The problem as I understand it is the "fewer than N (N>0)" scenario.

zenspider:
This works, but I dunno if you'll like it:

```diff
diff -r old/lib/minitest/mock.rb new/lib/minitest/mock.rb
--- old/lib/minitest/mock.rb
+++ new/lib/minitest/mock.rb
@@ -86,7 +86,8 @@
       when Hash then
         "#{name}(#{data[:args].inspect[1..-2]}) => #{data[:retval].inspect}"
       else
-        data.map { |d| __call name, d }.join ", "
+        s = data.map { |d| __call name, d }.join ", "
+        data.size < 2 ? s : "[ #{s} ]"
       end
     end
 
@@ -97,14 +98,11 @@
 
     def verify
       @expected_calls.each do |name, calls|
-        calls.each do |expected|
-          raise MockExpectationError, "expected #{__call name, expected}, got [#{__call name, @actual_calls[name]}]" if
-            @actual_calls.key?(name) and
-            not @actual_calls[name].include?(expected)
-
-          raise MockExpectationError, "expected #{__call name, expected}" unless
-            @actual_calls.key?(name) and
-            @actual_calls[name].include?(expected)
+        act = @actual_calls[name]
+        if calls != act then
+          act = __call name, act
+          act = ", got #{act}" unless act.empty?
+          raise MockExpectationError, "expected #{__call name, calls}#{act}"
         end
       end
       true
diff -r old/test/minitest/test_minitest_mock.rb new/test/minitest/test_minitest_mock.rb
--- old/test/minitest/test_minitest_mock.rb
+++ new/test/minitest/test_minitest_mock.rb
@@ -235,7 +235,7 @@
 
     e = assert_raises(MockExpectationError) { mock.verify }
 
-    exp = "expected foo(:baz) => nil, got [foo(:bar) => nil]"
+    exp = "expected [ foo(:bar) => nil, foo(:baz) => nil ], got foo(:bar) => nil"
 
     assert_equal exp, e.message
   end
@@ -343,6 +343,36 @@
     mock.verify
   end
 
+  def test_mock_was_called_x_times
+    mock = Minitest::Mock.new
+
+    mock.expect :foo, true, []
+    mock.expect :foo, true, []
+
+    mock.foo
+    mock.foo
+
+    mock.verify
+  end
+
+  def test_mock_was_never_called
+    mock = Minitest::Mock.new
+
+    mock.expect(:foo, true, [])
+
+    util_verify_bad "expected foo() => nil"
+  end
+
+  def test_mock_blow_up_when_called_fewer_times
+    @mock = Minitest::Mock.new
+
+    @mock.expect :foo, true, []
+    @mock.expect :foo, true, []
+
+    @mock.foo
+
+    util_verify_bad "expected [ foo() => true, foo() => true ], got foo() => true"
+  end
 end
 
 require "minitest/metametameta"
```

zenspider:
This is probably broken for the non-args branch.

zamith:
The thing is I want the was never called to be verified as correct. That would happen on a test where you want to make sure something doesn't happen, like for instance that a DB connection is never created. 

I can assert that it raises an exception, but that kind of defeats the it's purpose in the first place. Thus the times argument. However, I'd be good with having some specific API for expecting never to be called only, because again, I believe that's the only use case that is not possible with this approach.

Maybe `mock.expect_never_called(:foo, ret_val, args)`?

zenspider:
https://github.com/seattlerb/minitest/blob/master/test/minitest/test_minitest_mock.rb#L148-L156 ?

zamith:
Sorry, but I'm not sure I follow. That error is fine, I do want to get an error when that happens, but sometimes I also want to say that I want to make sure that a method is not called, without raising an error.

Like:

```
def test_mock_was_never_called
  mock = Minitest::Mock.new
  mock.expect_never_called(:foo, true, [])

  mock.verify
end

=> 1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

Does this make sense to you at all?

zenspider:
@zamith what timezone are you in? UTC-2 or so? Is there enough time overlap that we can go realtime on IRC or something?

zamith:
I'm UTC+1 right now. Sure, we I'll DM you on twitter, with contact details

zamith:
@zenspider Ended up sending you an email to the email on your github profile

pithub:
Hi @zamith, if you want to verify that a method isn't called, isn't it enough to "specify" no expect at all?

```ruby
  def test_verifies_no_method_is_called
    mock = Minitest::Mock.new

    # some code never calling a method on mock
    assert mock.verify # ==> passes

    assert_raises NoMethodError do
      mock.foo # ==> blows up
    end
  end
```

zenspider:
Sorry, I lost track of this. Yes, that was my point of confusion exactly.

zamith:
@pithub I meant to verify that a specific method is not called on the mock. Let's say you have a `DatabaseAdapter` class, and in one test you want to make sure that it doesn't reach to the db and doesn't call the `create_connection` method, although other methods can be called on the mock. 

I don't see that your example fits as a solution for this problem, or am I wrong?

pithub:
@zamith I think I still don't understand exactly what you need. When you say "other methods can be called on the mock", do you mean methods that have been `#expect`ed on the mock? I'm asking  because currently a `Minitest::Mock` accepts only method calls that have been `#expect`ed previously. All other method calls raise errors.

If you could write a little code for your DatabaseAdapter example, maybe this would help?

zamith:
@pithub I want to do something like this:

```ruby
def test_mock_was_never_called
  mock = Minitest::Mock.new
  mock.expect_never_called(:foo, true, [])

  mock.verify
end
```

And this should pass. The RSpec equivalent would be to use spies:

```ruby
it "never creates a connection" do
  database = spy
  database.other_method
  expect(database).not_to have_received(:create_connection)
end
```

zenspider:
@zamith ok. Let's work this out. This is my confusion in a nutshell:

```ruby
def test_mock_was_never_called
  mock = Minitest::Mock.new
  # mock.expect_never_called(:foo, true, [])

  mock.verify # passes
end
```

and

```ruby
def test_mock_was_never_called
  mock = Minitest::Mock.new
  
  mock.should_never_be_called # blows up
end
```

Help!

zamith:
Oh, I get your point now. It totally works that way. It is not, imo, super clear that by just creating a mock and calling `verify` on it, I am expecting no methods to be called on it, for someone reading the test. Even though it totally makes sense.

I'm closing this.

PR #591, opened 2015-10-20
Made a special exception handling for multiple stubs

mig-hub:
@zamith, here I drafted a possible handling for the multiple stub exception.
Not sure if checking `metaclass.nil?` is the most future proof way to decide not to run the undef/def of the `ensure` block, but in the current state it works.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -1,4 +1,5 @@
 class MockExpectationError < StandardError; end # :nodoc:
+class MultipleStubError < StandardError; end # :nodoc:
 
 module Minitest # :nodoc:
 
@@ -198,6 +199,10 @@ class Object
   def stub name, val_or_callable, *block_args
     new_name = "__minitest_stub__#{name}"
 
+    if respond_to? new_name
+      raise MultipleStubError, "method :%s was already stubbed" % [name.to_s]
+    end
+
     metaclass = class << self; self; end
 
     if respond_to? name and not methods.map(&:to_s).include? name.to_s then
@@ -222,8 +227,10 @@ class Object
 
     yield self
   ensure
-    metaclass.send :undef_method, name
-    metaclass.send :alias_method, name, new_name
-    metaclass.send :undef_method, new_name
+    unless metaclass.nil?
+      metaclass.send :undef_method, name
+      metaclass.send :alias_method, name, new_name
+      metaclass.send :undef_method, new_name
+    end
   end
 end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -200,7 +200,7 @@ class Object
     new_name = "__minitest_stub__#{name}"
 
     if respond_to? new_name
-      raise StubError, "method :%s was already stubbed" % [name.to_s]
+      raise StubError, "method :#{name.to_s} was already stubbed"
     end
 
     metaclass = class << self; self; end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -200,7 +200,7 @@ class Object
     new_name = "__minitest_stub__#{name}"
 
     if respond_to? new_name
-      raise StubError, "method :#{name.to_s} was already stubbed"
+      raise StubError, "method :#{name} was already stubbed"
     end
 
     metaclass = class << self; self; end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -199,7 +199,8 @@ class Object
   def stub name, val_or_callable, *block_args
     new_name = "__minitest_stub__#{name}"
 
-    if respond_to? new_name
+    restubbed = respond_to? new_name
+    if restubbed
       raise StubError, "method :#{name} was already stubbed"
     end
 
@@ -227,7 +228,7 @@ class Object
 
     yield self
   ensure
-    unless metaclass.nil?
+    unless restubbed
       metaclass.send :undef_method, name
       metaclass.send :alias_method, name, new_name
       metaclass.send :undef_method, new_name
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -1,5 +1,5 @@
 class MockExpectationError < StandardError; end # :nodoc:
-class MultipleStubError < StandardError; end # :nodoc:
+class StubError < StandardError; end # :nodoc:
 
 module Minitest # :nodoc:
 
@@ -200,7 +200,7 @@ class Object
     new_name = "__minitest_stub__#{name}"
 
     if respond_to? new_name
-      raise MultipleStubError, "method :%s was already stubbed" % [name.to_s]
+      raise StubError, "method :%s was already stubbed" % [name.to_s]
     end
 
     metaclass = class << self; self; end
--- a/test/minitest/test_minitest_mock.rb
+++ b/test/minitest/test_minitest_mock.rb
@@ -495,4 +495,18 @@ class TestMinitestStub < Minitest::Test
     @tc.assert_equal true, rs
   end
 
+  def test_blow_up_on_multiple_stub
+    h = {}
+    e = assert_raises MultipleStubError do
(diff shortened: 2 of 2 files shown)


lib/minitest/mock.rb
```diff
@@ -1,4 +1,5 @@
 class MockExpectationError < StandardError; end # :nodoc:
+class MultipleStubError < StandardError; end # :nodoc:
```
zamith: What about a more generic `StubError`?

lib/minitest/mock.rb
```diff
@@ -198,6 +199,10 @@ class Object
   def stub name, val_or_callable, *block_args
     new_name = "__minitest_stub__#{name}"
 
+    if respond_to? new_name
+      raise MultipleStubError, "method :%s was already stubbed" % [name.to_s]
```
zamith: Any particular reason not to use interpolation?

lib/minitest/mock.rb
```diff
@@ -222,8 +227,10 @@ def stub name, val_or_callable, *block_args
 
     yield self
   ensure
-    metaclass.send :undef_method, name
-    metaclass.send :alias_method, name, new_name
-    metaclass.send :undef_method, new_name
+    unless metaclass.nil?
```
zamith: I'm also not a big fan of this, but can't really think of a better way of doing it. @zenspider, ideas?

lib/minitest/mock.rb
mig-hub: Yeah true, the message gives the details.
I'll update.

lib/minitest/mock.rb
mig-hub: No at all. I wanted to stay as close as possible to the current coding style in many ways and saw it done this way up the page. I assumed it was for speed since this is (or used to be faster), although speed during the raising of an error is probably not key.

I didn't give it much thought, but I am happy to use interpolation instead.

lib/minitest/mock.rb
mig-hub: Indeed, I thought about saving the boolean as well. Like this:

```ruby
restubbed = respond_to? new_name
if restubbed
  # Raise
end
```

And then use this variable as well in the `ensure` block.
It is not much cleaner, but it is explicit and more future proof.

What do you think?

mig-hub:
@zamith , I have pushed the modifications you mentioned. Let me know if you prefer the other solution for the `ensure` block.

lib/minitest/mock.rb
```diff
@@ -198,6 +199,10 @@ class Object
   def stub name, val_or_callable, *block_args
     new_name = "__minitest_stub__#{name}"
 
+    if respond_to? new_name
+      raise StubError, "method :#{name.to_s} was already stubbed"
```
zamith: If I'm not mistaken the interpolation already calls the `to_s` method. Was this not the case in your tests?

lib/minitest/mock.rb
zamith: It seems a bit odd to have a conditional in an `ensure` block, but at least this is better than depending on the `metaclass` being defined or not.

zamith: I'd say go with this until (if) we can find a better solution.

lib/minitest/mock.rb
mig-hub: You are right.
Just a silly mistake sorry.
I fixed it.

lib/minitest/mock.rb
mig-hub: I could be wrong because I am not entirely sure the `ensure` block is designed only for exceptions that we predict when we write the code —in which case it is fine—, or to make sure it remains silent and harmless, even for unpredicted exceptions.

This is not easy in our case, because most `ensure` block I see tend to undo what the method was doing because it was not complete. Whereas this one is trying to enforce parts of what the method is supposed to do and contains a few things which can break if the exception is not a predictable one.

I'll have a look at the code to have a better understanding of why the `ensure` block is there in the first place and think about it.

mig-hub: @zamith, forget what I said about the "Undo", It is exactly what the current `ensure` block is doing.

I actually can make the tests pass with this at the end instead of `ensure`:

```ruby
out = yield self
metaclass.send :undef_method, name
metaclass.send :alias_method, name, new_name
metaclass.send :undef_method, new_name
out
```

Although it made me realise that while this passes, it just means that we can only test for what we predict. The `ensure` block is definitely a reassurance, although we have no real test which requires the `ensure` block.

Now I'm gonna **ensure** I have a cup of coffee.

zamith:
It looks good to me. /cc @zenspider

test/minitest/test_minitest_mock.rb
```diff
@@ -495,4 +495,18 @@ def test_mock_with_yield
     @tc.assert_equal true, rs
   end
 
+  def test_blow_up_on_multiple_stub
+    h = {}
+    e = assert_raises StubError do
+      h.stub(:empty?, false) do
+        h.stub(:empty?, true) do
```
zenspider: I have to ask: why would you have such a scenario? Even in rspec this doesn't make sense.

zamith: You could have a global stub, and for a specific method stub it in a different way. Would probably be a pain to understand those tests, but I can see that happening. 

However, I'd definitely not do that with Minitest, and would strongly encourage against it.

mig-hub: @zenspider The real scenario was with arguments which makes a bit more sense. I agree that here it looks like it is just reversing the previous stub. 

Maybe my test is not good enough because it does not show a believable scenario. Do you think I should change the test and use arguments?

Now possible scenarios are:

RSpec doesn't do it with blocs (afaik) but stubs are a one off, one call. So if you expect 2 calls with different arguments, in RSpec you have to stubs. They also have to happen in the same order. So the most mindless translation would be the nested stubs in Minitest. Let say you are building a converter, this is not such a bad exception to raise because you see immediately what is wrong.

But if you go further, one can create a method which hides the stubbing and yield in her/his tests. In which case the stubs are not on top of each other and can be missed.

zamith: Do you have a concrete example of that situation? I feel you should be able to work around it by narrowing the scope of what you're testing in each test, through dependency injection and mocking, or some other testing/design practice that will make the use of nested stubs like that, pointless.

In any case, I'm almost certain you can find a refactor that will make it easier to test that situation.

mig-hub: There has never been a doubt that this is bad code. I thought the question was "Do we make a specific error message or not?". I didn't mean scenarios where somebody would need this functionality. I meant scenarios where this mistake could happen.

The confusion probably comes from "my test is not good enough". I meant the test for the exceptions. I don't doubt that the nested stub is bad.

And then that is what exceptions are for, otherwise there would not be any "ArgumentError: wrong number of arguments" in ruby.

zamith: Right. I think where @zenspider is trying to get is, to understand whether this is a bad enough practice that we don't even have an error for it.

mig-hub: Okay I wasn't sure. Yeah I'm happy to drop it since it is not gonna make a big improvement on Minitest. I would rather do something which is useful.

Although "rare enough" would be a better reason for me than "bad enough". Using the wrong number of arguments on a method is "bad", but unfortunately "ArgumentError: wrong number of arguments" is not that rare. Exceptions rarely show when your code is "good", that is the point, isn't it?

I reckon it is rare enough to drop the pull request.

zamith: Sure, rare enough is probably a better wording.

zenspider:
@zamith I still don't understand this problem so I'm not tracking it anymore. Your call.

zamith:
@zenspider I think this is too rare of an issue for us to handle it, at this point. If it does come up more often we can always revisit this.

zenspider:
Closing for now.

Issue #582, opened 2015-10-04
Calling mock in other thread


panitaxx:
Hello. 

When calling a mock from a thread exceptions get lost and mock.verify does not works (it asserts true). Is there any solution for this. Here is an example 

```ruby
require "minitest/autorun"

class C
  def initialize(pr)
    @pr=pr
  end

  def run(a,b)
    Thread.new(a,b,&@pr)
  end
end

class TestC  < Minitest::Test

  def test_multi_call
    pr_mock = MiniTest::Mock.new
    pr_mock.expect(:test,nil,[1,2])
    pr = Proc.new { |a,b| pr_mock.test(1,2) }
    cinst = C.new(pr)
    cinst.run(1,2)
    sleep(0.5)
    cinst.run(1,2)
    assert(pr_mock.verify)
  end
end
```

zamith:
Do you have a real world scenario where this would be useful? This is an interesting problem, just unsure of whether it is something Minitest should resolve or not.

panitaxx:
I have this "alarm" system where an alarm starts a thread. Each alarm has different procs. I just wanted to test out that the proc really gets called and gets called only once (since each alarm can repeat lots of times but the thread should be started only once). 

My workaround was to catch the exception in the  `Proc` and exit right there. Maybe just put some guidance on the readme so people knows how it works with threads.

zamith:
Would testing side effects of that suffice?

Adding something on the README I think is a good idea regardless of whether we fix this or not. @zenspider what do you think? Should this be fixed, or taken as a feature and explained on the README, only?

zenspider:
This code example isn't sound. You're spawning threads and throwing them away before verifying that they've finished running.

This is partially how threads work by default. See [Thread exception handling](http://docs.ruby-lang.org/en/2.2.0/Thread.html#class-Thread-label-Exception+handling) for more information.

zenspider:
@zamith your call. Personally, I think the examples given are not good examples, so they're not valid reasons to force change. Want to see that a proc was called? Set a boolean:

```ruby
  def test_multi_call
    called = false
    pr = Proc.new { called = true }
    cinst = C.new(pr)
    cinst.run(1,2)
    sleep(0.5) # this seems bad
    cinst.run(1,2) # this seems bad, and unexplained
    assert called, "proc not called"
  end
```

zamith:
[...]

PR #596, opened 2015-11-16
Add a note to the README to inform about mocks and threads

zamith:
Why:

* In response to #582

* Adding an example of how to test mocks in a multi-threaded
environment

@zenspider I'm not sure of the tone, but in terms of info I'd say it's there. What do you think?


zenspider:
Done. Thanks!

PR #597, opened 2015-11-26
Add minitest gems info

kematzy:
Hi, 

I have created three Minitest gems that you hopefully want to include in the README. I added each gem as a separate commit just in case you want to pick and choose.

Hope this is OK.

PS. It would be great if the list of Minitest gems had actual links to the gems. Would you consider a separate pull request with actual links to most (all?) of the gems referred to?

Thanks for your time and attention.


zenspider:
Done. Thanks!

zenspider:
No, I don't want full links to the gems. It'd make the readme unreadable.

PR #599, opened 2015-12-04
Add assert_raises_with_message

brianstorti:
This adds support to `assert_raises_with_message`. The usage is straightforward:
```ruby
assert_raises_with_message RuntimeError, "expected error message"
```

and in case of failure, it gives us a message like this:

```
RuntimeError raised with unexpected message
Expected: "blah"
  Actual: "foo".
```

It relies on `assert_raises`, so if no exception or an unexpected exception is raised, the behavior is the same as `assert_raises`.

Would love some feedback to know if this is a feature you'd like to see included and if the implementation looks good.

Thank you.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -318,6 +318,19 @@ module Minitest
     end
 
     ##
+    # Fails unless the block raises +err+ with the message +err_msg+.
+
+    def assert_raises_with_message err, err_msg, msg = nil
+      e = assert_raises(err, msg) { yield }
+      if err_msg != e.message
+        flunk message(msg) {
+          diff_message = diff err_msg, e.message
+          "#{e.class} raised with unexpected message\n#{diff_message}"
+        }
+      end
+    end
+
+    ##
     # Fails unless +obj+ responds to +meth+.
 
     def assert_respond_to obj, meth, msg = nil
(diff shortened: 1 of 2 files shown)


lib/minitest/assertions.rb
```diff
@@ -318,6 +318,19 @@ def assert_raises *exp
...
+    # Fails unless the block raises +err+ with the message +err_msg+.
+
+    def assert_raises_with_message err, err_msg, msg = nil
+      e = assert_raises(err, msg) { yield }
+      if err_msg != e.message
+        flunk message(msg) {
+          diff_message = diff err_msg, e.message
+          "#{e.class} raised with unexpected message\n#{diff_message}"
+        }
+      end
```
zenspider: Why not use `assert_equal`?

phiggins: Or even `assert_match`, in case you only care about some pattern in the exception message.

brianstorti: That was the initial idea, but the error message was not very clear using just `assert_equal`. Do you think that'd be enough?

zenspider:
After more thought... I think I'm going to have to respectfully decline. This assertion is doing two things and that just feels like too much to me. I'm happy to help you publish this as a separate gem if you'd like to get it out there.

brianstorti:
That's alright. Thanks for taking the time to review it.

PR #600, opened 2015-12-18
Remove outdated documentation

ivantsepp:
Remove outdated documentation since `parallel_each.rb` doesn't exist anymore and `run_one_method` is the method that `parallel.rb` overrides.


zenspider:
Good catch! Thanks!

PR #603, opened 2016-01-08
add negative filter by test name #519.

utilum:


zenspider:
Merged. Thank you so much. Can you please re-submit the filtering by file so it is open and discussable?

PR #614, opened 2016-03-31
Customizable signal for dumping info to STDERR

nate:
https://github.com/seattlerb/minitest/issues/560

### DIFF
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -18,12 +18,6 @@ module Minitest
     self.io_lock = Mutex.new
 
     ##
-    # The signal to use for dumping information to STDERR. Defaults to "INFO".
-
-    class << self; attr_accessor :info_signal; end
-    self.info_signal = "INFO"
-
-    ##
     # Call this at the top of your tests when you absolutely
     # positively need to have ordered tests. In doing so, you're
     # admitting that you suck and your tests are weak.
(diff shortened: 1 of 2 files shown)


lib/minitest/test.rb
```diff
@@ -18,6 +18,12 @@ class << self; attr_accessor :io_lock; end # :nodoc:
     self.io_lock = Mutex.new
 
     ##
+    # The signal to use for dumping information to STDERR. Defaults to "INFO".
+
+    class << self; attr_accessor :info_signal; end
+    self.info_signal = "INFO"
+
```
zenspider: Why in two places?

nate: I wasn't sure where to put it and was hoping for feedback. Where would you prefer?

zenspider: Drunk at ruby on ales. I'll get back to you when sober.

zenspider: Using `Minitest.info_signal` inside of minitest/test.rb should be sufficient.

nate:
I think this is good, could you take another look?

nate:
Hmmm, I just noticed you said minitest/test.rb, but I think it's a better fit in minitest.rb since there is already some top-level configuration setup there.

zenspider:
You have identical definitions in both minitest.rb and minitest/test.rb. I was trying to suggest(?)/suggesting keeping the one in minitest.rb and having minitest/test.rb refer to the former.

zenspider:
Done. Thanks!

PR #615, opened 2016-04-06
Pluralize final summary report line

richardtemple:
This is my first real pull request, so please forgive me if I'm doing it wrong.  I just went to Ruby on Ales and was inspired by the talks (including zenspider's) to finally give contributing a go.  I wasn't sure where to put the utility method `#simple_pluralize`, so I left it under `#summary` for now.  I also see that there are some other areas where pluralization could be used, but I wanted to run this by seattlerb first before continuing incase there are good reasons to not pluralize reporting.  Also, any advice in helping me become a better contributor would be much appreciated.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -620,11 +620,9 @@ module Minitest
          extra]
     end
 
-    def simple_pluralize(n, singular, plural=nil)
+    def simple_pluralize(n, singular)
       if n == 1
           "1 #{singular}"
-      elsif plural
-          "#{n} #{plural}"
       else
           "#{n} #{singular}s"
       end
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -610,9 +610,26 @@ module Minitest
       extra = "\n\nYou have skipped tests. Run with --verbose for details." if
         results.any?(&:skipped?) unless options[:verbose] or ENV["MT_NO_SKIP_MSG"]
 
-      "%d runs, %d assertions, %d failures, %d errors, %d skips%s" %
-        [count, assertions, failures, errors, skips, extra]
+      # list counts for runs, assertions, failures, errors, skips, and extra
+      "%s, %s, %s, %s, %s%s" %
+        [simple_pluralize(count, "run"),
+         simple_pluralize(assertions, "assertion"),
+         simple_pluralize(failures, "failure"),
+         simple_pluralize(errors, "error"),
+         simple_pluralize(skips, "skip"),
+         extra]
     end
+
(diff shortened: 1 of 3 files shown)


lib/minitest.rb
```diff
@@ -610,9 +610,26 @@ def summary # :nodoc:
...
+         simple_pluralize(errors, "error"),
+         simple_pluralize(skips, "skip"),
+         extra]
+    end
+
+    def simple_pluralize(n, singular, plural=nil)
+      if n == 1
+          "1 #{singular}"
+      elsif plural
+          "#{n} #{plural}"
```
zenspider: why is this section here if you don't use it?

zenspider:
The main reason why I didn't do this already was just to maintain simplicity. I don't see anything overtly bad about this PR (tho I did add a q to the commit) but I don't know if I want to accept it for the reason mentioned.

zenspider:
Where else do you think this should be applied?

lib/minitest.rb
richardtemple: Do you mean the optional plural argument?  I saw it in another example and thought it made sense  if needed later.  I could take it out of course.

richardtemple:
There was just one other place above in the `statistics` method.

richardtemple:
Unfortunately, I'm not familiar with adding a q to a commit.

lib/minitest.rb
zenspider: Yeah. The whole philosophy behind minitest is that "mini" part. It doesn't have any code it doesn't need. IOW: "If needed later" == "it isn't needed"

zenspider:
You mean here: https://github.com/richardtemple/minitest/blob/master/lib/minitest.rb#L587

richardtemple:
Cool github feature.  Yes, that was the method.  Thanks for being patient with my newbness.  (I've been using CVS and subversion doing Java and C# for a number of years.)

lib/minitest.rb
richardtemple: Sounds extremely reasonable.  :)  I'll commit the change in my fork incase you're on the fence about merging it.

zenspider:
To be fair... I'm still on the fence even with that cleanup.

richardtemple:
Well, as much as I'd love to have my first accepted PR be for something as great as minitest, if this doesn't cut it, I've still learned more about working via github.  I really appreciate your candid comments.  Thank you!

zenspider:
Yeah. Thank you very much for the contribution, but I'm going to decline.

richardtemple:
Cool.  Thank you for considering.

PR #619, opened 2016-04-15
Remove Interrupt from PASSTHROUGH_EXCEPTIONS

waldyr:
As Interrupt is also a SignalException we don't need to explicitly put it into the array.

```
Interrupt.ancestors # => [Interrupt, SignalException, ... ]
```


zenspider:
Cool, thanks!

PR #623, opened 2016-05-09
Fix assert_throw rescuing any NameError and ArgumentError

### LINKED ISSUE #622: assert_throw returns confusing message if ArgumentError is thrown
If a normal ArgumentError is raised during a test assert_throws will catch it and a very confusing message is returned because of this code:

```
    rescue ArgumentError => e     # 1.9 exception
      default += ", not #{e.message.split(/ /).last}"
```

So you get the last part of the argument error returned as the name of a missing argument like:

`Expected :denied to have been thrown, not :id.`

waldyr:
Fix #622 

`throw` have different behavior depending on Ruby version:
- Ruby 1.8 raises `NameError` with `name` equal to the `throw`'s  `tag`
- Ruby 1.9 raises `ArgumentError` with the following `message`: `uncaught throw #{throw's tag}`
- Ruby 2.2 raises `UncaughtThrowError` which is also an `ArgumentError` and is handled accordingly.

So the simplest solution is to re-raise the exception in case of a different source other than the `throw`

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -386,8 +386,10 @@ module Minitest
         rescue ThreadError => e       # wtf?!? 1.8 + threads == suck
           default += ", not \:#{e.message[/uncaught throw \`(\w+?)\'/, 1]}"
         rescue ArgumentError => e     # 1.9 exception
+          raise e unless e.message.include?("uncaught throw")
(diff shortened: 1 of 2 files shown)


zenspider:
I like this. Merged. Thanks!

Issue #530, opened 2015-04-26
Interest in assert_raises_with_message?


fny:
`TestUnit` provides an [`assert_raise_with_message` assertion](http://ruby-doc.org/stdlib-2.1.0/libdoc/test/unit/rdoc/Test/Unit/Assertions.html#method-i-assert_raise_with_message).  Is `assert_raises_with_message` considered in scope for Minitest?

Right now it's not terribly painful take care manually, but it definitely becomes a bit cumbersome when testing a lot of exceptions:

``` ruby
 err = assert_raises(NameError) { blah }
 assert_equal "NameError: undefined local variable or method `blah'", err.message
```

Rough implementation:

``` ruby
def assert_raises_with_message(exception, err_message, msg = nil)
  err = assert_raises(exception, msg) { yield }
  if err_message.is_a?(Regexp)
    assert_matches err_message, err.message
  else
    assert_equal err_message, err.message
  end  
end
```

zenspider:
Not sure about implementation, but yes, interested.

fny:
In the tests, there's a check in the teardown that ensures only one assertion is fired. It seems to make sense for two assertions (assert_raises then assert_equal/assert_matches) to be called in this case. Thoughts?

```ruby
def assert_raises_with_message(exception, expected_msg, msg = nil)
  case expected_msg
  when String
    assert = :assert_equal
  when Regexp
    assert = :assert_match
  else
    raise TypeError, "Expected #{mu_pp(expected_msg)} to be a kind of String or Regexp, not #{expected_msg.class}"
  end
  ex = assert_raises(exception, msg) { yield }
  msg = message(msg, "") {"Expected Exception (#{exception}) to be raised, but the message doesn't match"}

  if assert == :assert_equal
    assert_equal(expected_msg, ex.message, msg)
  else
    msg = message(msg) { "Expected #{mu_pp(expected_msg)} to match #{mu_pp(ex.message)}" }
    assert_matches expected_msg, ex.message, msg
  end
  ex
end
```

zenspider:
I'm really not fond of that level of complexity. Let me think about an alternative.

AlexWayfer:
I'm very interested in this assertion.

knightq:
Me too!

zenspider:
I'm not interested in this at this point.

I have modified assert_raises to return the exception

PR #627, opened 2016-05-22
Add assertions assert_raises_with_message

waldyr:
#530

`assert_raises_with_message`:
- handles exceptions exactly the same way `assert_raises` does. (And can be dried up if you think it's necessary)
- `assert_equals` in the exception's message

What do you think about the complexity of it?

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -338,24 +338,12 @@ module Minitest
     # passed.
 
     def assert_raises_with_message exp, exp_msg, msg = nil
-      begin
+      e = assert_raises exp do
         yield
-      rescue exp => e
-        return assert_equal exp_msg, e.message, msg
-      rescue Minitest::Skip, Minitest::Assertion
-        # don't count assertion
-        raise
-      rescue SignalException, SystemExit
-        raise
-      rescue Exception => e
-        flunk proc {
-          exception_details(e, "#{msg}#{mu_pp(exp)} exception expected, not")
-        }
       end
 
-      flunk "#{msg}#{mu_pp(exp)} expected but nothing was raised."
+      assert_match Regexp.new(exp_msg), e.message, msg
     end
-
 
     ##
     # Fails unless +obj+ responds to +meth+.
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -331,6 +331,33 @@ module Minitest
     end
 
     ##
+    # Fails unless the block raises +exp+ and +exp+'s message is equal to +msg+.
+    #
+    # +exp+ takes an optional message on the end to help explain
+    # failures and defaults to StandardError if no exception class is
+    # passed.
+
+    def assert_raises_with_message exp, exp_msg, msg = nil
+      begin
+        yield
+      rescue exp => e
+        return assert_equal exp_msg, e.message, msg
+      rescue Minitest::Skip, Minitest::Assertion
+        # don't count assertion
+        raise
+      rescue SignalException, SystemExit
+        raise
+      rescue Exception => e
+        flunk proc {
+          exception_details(e, "#{msg}#{mu_pp(exp)} exception expected, not")
+        }
+      end
+
+      flunk "#{msg}#{mu_pp(exp)} expected but nothing was raised."
+    end
+
+
+    ##
     # Fails unless +obj+ responds to +meth+.
 
     def assert_respond_to obj, meth, msg = nil
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -1356,6 +1356,34 @@ class TestMinitestUnitTestCase < Minitest::Test
     end
   end
 
+  def test_assert_raises_with_message_wrong_message
+    e = assert_raises Minitest::Assertion do
+      @tc.assert_raises_with_message StandardError, "test" do
+        raise "txt"
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


tenderlove:
Why not this:

```diff
diff --git a/lib/minitest/assertions.rb b/lib/minitest/assertions.rb
index 54f7b88..c8c6680 100644
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -331,6 +331,21 @@ module Minitest
     end
 
     ##
+    # Fails unless the block raises +exp+ and +exp+'s message is equal to +msg+.
+    #
+    # +exp+ takes an optional message on the end to help explain
+    # failures and defaults to StandardError if no exception class is
+    # passed.
+
+    def assert_raises_with_message exp, exp_msg, msg = nil
+      e = assert_raises exp do
+        yield
+      end
+      assert_equal exp_msg, e.message, msg
+    end
+
+
+    ##
     # Fails unless +obj+ responds to +meth+.
 
     def assert_respond_to obj, meth, msg = nil
diff --git a/test/minitest/test_minitest_test.rb b/test/minitest/test_minitest_test.rb
index dcc7ed1..122f440 100644
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -1356,6 +1356,35 @@ class TestMinitestUnitTestCase < Minitest::Test
     end
   end
 
+  def test_assert_raises_with_message_wrong_message
+    e = assert_raises Minitest::Assertion do
+      @tc.assert_raises_with_message StandardError, "test" do
+        raise "txt"
+      end
+    end
+
+    expected = "Expected: \"test\"\n  Actual: \"txt\""
+    assert_equal expected, e.message
+  end
+
+  def test_assert_raises_with_message_wrong_message
+    e = assert_raises Minitest::Assertion do
+      @tc.assert_raises_with_message StandardError, "test" do
+        # do not raise
+      end
+    end
+
+    expected = "StandardError expected but nothing was raised."
+    assert_equal expected, e.message
+  end
+
+  def test_assert_raises_with_message
+    @assertion_count += 1
+    @tc.assert_raises_with_message StandardError, "test" do
+      raise "test"
+    end
+  end
+
   def test_assert_raises_subclass_triggered
     e = assert_raises Minitest::Assertion do
       @tc.assert_raises SomeError do
```

waldyr:
I cannot believe I'm answering @tenderlove. I would never imagine establishing any sort of communication with you hahaha. I'm a huge fan of yours.

A while after I've opened the PR I thought on doing `assert_raises_with_message` like your code but I had some reasons not to.
- Your code is how most people is used to test the exception's message. This probably invalidates the need of having such an assertion.
- `assert_raises_with_message` is an assertion (method) with two responsibilities.
- `assert_raises_with_message` with two assertions, it will messes the expected one-to-one assertion count (unless I somehow fix it).

So before doing anything more I waited for guidance. I also thought in probably change that `assert_equal` into one `assert_match`. What do you guys think?

PS.: send my regards to Gorbachev and Choo Choo

nateberkopec:
@waldyr Eh, don't sweat it. I've met him IRL and he smells kinda funny.

waldyr:
hahahaha Is this real life? Are you guys doing this on purpose? Just to make a random contributor guy happy? I'm speechless. 😁 😁 😁 😁 

One of my friends recently borrowed me your book (The Complete Guide to Rails Performance). A moment ago I was reading it and now I'm talking to you hahahaha. That's outstanding. I'll definitely contribute to minitest whenever possible. 

Very nice talking to you too @nateberkopec. Your RubyConf 2016's presentation is awesome!!!

PS.: Thank you so much @zenspider. Your Mountain West Ruby 2016's presentation caught my eye for minitest and delivered this awesome moment.

blowmage:
> `assert_raises_with_message` is an assertion (method) with two responsibilities

Yes, but isn't adding expected message to `assert_raises` is also adding a second responsibility?

> messes the expected one-to-one assertion count

Is that a real life concern? Honest question. Other assertions make multiple asserts. (`assert_empty`, `assert_includes`, etc...)

waldyr:
> Is that a real life concern? Honest question. Other assertions make multiple asserts. (assert_empty, assert_includes, etc...)

I did not know that. If some assertions is already been doing multiple asserts I'm going to value readability and therefore update my solution closer to @tenderlove's solution.

> Yes, but isn't adding expected message to assert_raises also adding a second responsibility?

Sorry but I didn't get it. Could you please rephrase it?

I've also added one more test case to work with regular expressions.

blowmage:
> Could you please rephrase it?

My bad. I thought that your proposal was to change the `assert_raises` signature from `assert_raises exp, msg = nil` to `assert_raises exp, exp_msg, msg = nil`. I must have been looking at some other code about this.

The point I was trying to make is that there are two different checks happening here. I understand the desire to check both in a single assertion, but I honestly prefer to make [two assertions](http://stackoverflow.com/a/14428915/5627) instead of the one. I think it reads and communicates intent better. Just my opinion though.

waldyr:
> I understand the desire to check both in a single assertion, but I honestly prefer to make two assertions instead of the one.

I've seen #530 and an apparent interest from Ryan. Your point is one of the reasons that I waited after opening the PR. I didn't know if this would end well. Let's see what happens.

Nakilon:
I came here to say that I want this feature too, but the fact that `assert_raises` returns the exception and the link just above my comment tells me that this feature is really not needed.

tonytonyjan:
I think we don't need this. Testing error message makes little sense to me.

PR #630, opened 2016-06-23
Bignum and Fixnum have been merged to Integer

tenderlove:
On Ruby 2.4 Bignum and Fixnum have been merged in to Integer, so some of
the test messages are different on Ruby 2.4.  This commit asks the class
for it's name so that the tests will work on all Ruby versions


utilum:
These tests do fail with ruby 2.4.0preview1, and pass with this fix.

tenderlove:
Looks like this was merged in fc96854d3bf6495ea9f7dad821bb64f6bf03440f

PR #634, opened 2016-08-04
Add line number in methods defined by "it"

tenderlove:
This commit adds the line number of the method defined by "it" blocks in
minitest/spec.  Our tests at work are "spec" style (they use the DSL to
define tests), and we add a line number to the method name that gets
defined.  The reason I like this is because I can write a test, then run
that single test without trying to guess how the test framework defines
the method, I just do:

```
$ ruby test_file.rb -n /L56/
```

This also makes running the single test from Vim quite easy because the
editor can just grab the line number and pass it as a parameter to the
test runner.

This patch only adds the line number if the Ruby version being used
supports `source_location` on the `Proc` object, so this should work on
all versions of Ruby.


zamith:
Have you considered using [m](https://github.com/qrush/m)? Not a fan of using a binary other than `ruby` to run the tests?

tenderlove:
@zamith I am quite happy with my existing dev setup, thanks.

zenspider:
MT6 added line support including multiple lines and/or ranges.

PR #636, opened 2016-08-04
Don't shuffle runnables unless :random or :parallel

chrisnicola:
Don't shuffle Runnable.runnables when running tests with test_order set to :random or :parallel.

I have not added any test for this as there doesn't seem to be any place where this part of the test runner is currently being tested.

I also considered it might make sense to encapsulate this logic in the `Runnable` class. This would probably be easier to write unit tests for as well.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -147,7 +147,13 @@ module Minitest
   # sub-classes to run.
 
   def self.__run reporter, options
-    suites = Runnable.runnables.shuffle
+    suites = case Minitest::Test.test_order
+    when :random, :parallel then
+      Runnable.runnables.shuffle
+    else
+      Runnable.runnables
+    end
+
     parallel, serial = suites.partition { |s| s.test_order == :parallel }
 
     # If we run the parallel tests before the serial tests, the parallel tests
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -158,7 +158,7 @@ module Minitest
     when :random then
       Runnable.runnables.shuffle
     else
-      Runnable.runnables.sort_by { |r| r.name }
+      Runnable.runnables
     end
 
     if Minitest.suite_order == :random
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -45,6 +45,12 @@ module Minitest
   mc.send :attr_accessor, :info_signal
   self.info_signal = "INFO"
 
+  SUITE_ORDERINGS = [:sorted, :random]
+
+  def self.suite_order
+    @@suite_order || :random
+  end
+
   ##
   # Registers Minitest to run at process exit
 
@@ -120,6 +126,7 @@ module Minitest
     self.load_plugins
 
     options = process_args args
+    @@suite_order = options[:order]
 
     reporter = CompositeReporter.new
     reporter << SummaryReporter.new(options[:io], options)
@@ -147,22 +154,26 @@ module Minitest
   # sub-classes to run.
 
   def self.__run reporter, options
-    suites = case Minitest::Test.test_order
-    when :random, :parallel then
+    suites = case suite_order
+    when :random then
       Runnable.runnables.shuffle
     else
-      Runnable.runnables
+      Runnable.runnables.sort_by { |r| r.name }
     end
 
-    parallel, serial = suites.partition { |s| s.test_order == :parallel }
+    if Minitest.suite_order == :random
+      parallel, serial = suites.partition { |s| s.test_order == :parallel }
 
-    # If we run the parallel tests before the serial tests, the parallel tests
-    # could run in parallel with the serial tests. This would be bad because
-    # the serial tests won't lock around Reporter#record. Run the serial tests
-    # first, so that after they complete, the parallel tests will lock when
-    # recording results.
-    serial.map { |suite| suite.run reporter, options } +
-      parallel.map { |suite| suite.run reporter, options }
+      # If we run the parallel tests before the serial tests, the parallel tests
+      # could run in parallel with the serial tests. This would be bad because
+      # the serial tests won't lock around Reporter#record. Run the serial tests
+      # first, so that after they complete, the parallel tests will lock when
+      # recording results.
+      serial.map { |suite| suite.run reporter, options } +
+        parallel.map { |suite| suite.run reporter, options }
+    else
+      suites.map { |suite| suite.run reporter, options }
+    end
   end
 
   def self.process_args args = [] # :nodoc:
@@ -195,6 +206,12 @@ module Minitest
 
       opts.on "-e", "--exclude PATTERN", "Exclude /regexp/ or string from run." do |a|
         options[:exclude] = a
+      end
+
+      opts.on "-o", "--order ORDER", "Order to run test suites in parallel, random, sorted." do |a|
+        order = a.to_sym
+        raise OptionParser::InvalidOption, "Order must be parallel, random, sorted or fixed" unless SUITE_ORDERINGS.include? order
+        options[:order] = order
       end
 
       unless extensions.empty?
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -22,10 +22,7 @@ module Minitest
     # admitting that you suck and your tests are weak.
 
     def self.i_suck_and_my_tests_are_order_dependent!
-      class << self
-        undef_method :test_order if method_defined? :test_order
-        define_method :test_order do :alpha end
-      end
+      @test_order = :alpha
     end
 
     ##
@@ -52,6 +49,11 @@ module Minitest
       extend Minitest::Parallel::Test::ClassMethods
     end
 
+    def self.run reporter, options
+      @test_order = options[:order] if options[:order]
+      super
+    end
+
     ##
     # Returns all instance methods starting with "test_". Based on
     # #test_order, the methods are either sorted, randomized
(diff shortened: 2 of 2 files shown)


chrisnicola:
@tenderlove I'm working on something that address your concern about setting this using the `Minitest::Test.test_order`. It is difficult to do in a way that I feel comfortable with, partly because `test_order` has been made so intentionally difficult to override.

zenspider:
Huh? test_order is just a method. It's the easiest thing to override. Please don't editorialize. It's not necessary.

chrisnicola:
There is no editorializing, but thanks for editing my comment to free it up from any possible criticism of the code. We wouldn't want reasonable discourse to occur.

Let me be more specific then. You are literally calling undef and def to change it with `i_suck_and_my_tests_are_order_dependent!` this is completely unnecessary ceremony and server to require the user to monkey patch the class if they want to change the value globally. On could argue nothing is hard to override in Ruby if that's what you are willing to do.

lib/minitest.rb
```diff
@@ -147,16 +154,26 @@ def self.run args = []
...
-    # first, so that after they complete, the parallel tests will lock when
-    # recording results.
-    serial.map { |suite| suite.run reporter, options } +
-      parallel.map { |suite| suite.run reporter, options }
+    suites = case suite_order
+    when :random then
+      Runnable.runnables.shuffle
+    else
+      Runnable.runnables.sort_by { |r| r.name }
+    end
```
tenderlove: I don't think sorting by name is the correct default because it will reorder people's tests.  Tests are added to `runnables` in load order, which isn't necessarily alphabetic (and if you're using dir globs then it will even change per platform).  If I configure minitest not to muck about with ordering, then I'd expect it to run the tests in the order I specified (especially if I wrote a script to require the tests in a particular order).

chrisnicola: Yeah I wasn't quite certain about that. I went back and forth, but I think it is fine to just not sort it at all.

I also noticed while try to support a `:fixed` option that Ruby actually sorts method names declared at the same context level itself so you can't list them in the order they were explicitly declared, which is interesting.

I'll change this to remove the sorting option altogether.

lib/minitest.rb
```diff
@@ -45,6 +45,12 @@ module Minitest
   mc.send :attr_accessor, :info_signal
   self.info_signal = "INFO"
 
+  SUITE_ORDERINGS = [:sorted, :random]
+
+  def self.suite_order
+    @@suite_order || :random
```
tenderlove: If you add `:order => :random` to the default `options` hash `process_args`, would it eliminate this conditional?

chrisnicola: Yes that's a good idea. I didn't see that.

nateberkopec:
> but since I've never been impacted by shuffled classes I've only ever twiddled this bit a class at a time (not for the base class).

That's interesting, because I've only used the base class setting. 

> undefine_method and define_method are currently used to do this when i_suck_and_my_tests_are_order_dependent! is called.

I'm pretty sure there is a reason for this, but I'll let Ryan defend his rationale. In any case, I think this change is not related to the original purpose of this PR (suite ordering), and should be removed and discussed in a separate PR.

> But I would argue that code like this should make easy affordances for configuration and extension rather than resisting it and trying to force users through a path of insulting themselves to achieve a simple goal.

I'm sorry you feel that way. Minitest is highly opinionated software, and in many cases deliberately makes things painful (mocks, stubs, etc). It is not a piece of software which *always* provides easy affordances for all possible uses. Minitest also has opinions about what "readable" and "extendable" code is, which often conflicts with more "mainstream" conceptions of Ruby. [I gave an entire conf talk about this.](http://confreaks.tv/videos/rubyconf2015-a-guided-read-of-minitest).

**TL:DR;** I am *for* the changes to `minitest.rb` and *against* the changes to `test.rb`.

chrisnicola:
> I'm sorry you feel that way.

I'm not exactly sure what this statement is supposed to mean. As far as I can tell I've simply made an case against providing APIs which lack appropriate affordances for using them in order to discourage their use. This is not an emotional appeal so I don't see how there is any need for you to apologize for how I feel.

> Minitest is highly opinionated software

This kind of statement seems common among Ruby projects. It seems to communicate that this is the end of the discussion so I will leave it here as well.

I think we all seem to agree with the main point that there is a bug here that should be squashed. And I am more than happy if you wish to accept the changes here in whole, in part, or not at all. For my part I've already affected a sufficient patch for my own purposes.

ntl:
If there is _no_ valid use case for wanting any other test ordering but random, then there should be no means of altering test ordering whatsoever. On the other hand, if there is _at least one_ valid use case -- even if it is just one among many invalid use cases -- then the library should assume that the user knows what she is doing and has a valid reason for specifying an alternative test order.

I get that being "opinionated" is a design goal of Minitest. That's valuable. Choosing random order as the default is the correct expression of that opinion. As long as Minitest offers the ability to specify a different test order, then that affordance should be offered with the same baseline level of usability as the rest of the library.

As a long time user of Minitest, I support both changes.

nateberkopec:
> But as long as Minitest offers the ability to specify a different test order,

Hah, don't tempt him ;)

nateberkopec:
> if there is at least one valid use case -- even if it is just one among many invalid use cases -- then the library should assume that the user knows what she is doing and has a valid reason for specifying an alternative test order.

FWIW, I think what Minitest assumes the *only* valid use-case is for non-random test ordering is for fixing bugs in your test suite. This is why (one of) the main way of changing it is called "i_suck_and...", because it's *intended* to be an ugly stain on your test suite that you should want to fix and remove. @chrisnicola mentioned on twitter that you might want non-random *reporting* of a random test run, but that's a separate issue.

As for your point, I guess I just disagree. Minitest, as written r/n, sees test ordering as only valid in a particular circumstance. I don't see why it has to provide affordances for all other possible circumstances.

lib/minitest.rb
tenderlove: It's probably better to return `nil` than the wrong value.  For example, someone may have specified "ordered" on the command line, but any method calling this method will assume that the test ordering is random until `run` is called.  If the method returns `nil`, at least callers will know that "it hasn't been set yet".  IOW, right now callers can't tell the difference between "someone specified `random`" and "test ordering hasn't been set yet".

sbellware:
Is it possible that the opinions and intentions of minitest don't reflect the highest achievements of test design technique?

But if it does, why is it necessary then to disparage students of the techniques because they are not at the highest level of achievement as expressed by minitest's opinions?

It strikes me as mere posturing to encode such harmful rhetoric as `I_suck_...` into the product itself, and I can't help but see insecurity in something that has to be expressed with such force and such contempt.

zenspider:
TL;DR: I'm done.

### I'm Tired of This

> But I would argue that code like this should make easy affordances for configuration and extension rather than resisting it and trying to force users through a path of insulting themselves to achieve a simple goal.

Can you stop with the editorializing already? [I've already asked once before](#issuecomment-237661784).

Nobody and nothing is forcing anyone to do anything.

```ruby
def self.test_order
  :alpha
end
```

See how self-insulting that is? See how you were forced into it? Wait, it isn't insulting and you weren't forced. It's just a method. And like any other method in ruby, you can override it. `Test.test_order` is even documented as such, so your editorializing is only meant to perturb... and it is working.

Your concern about my use of `undef_method` and `define_method` is misplaced. You assume bad intentions when there are none. If you write clean code and run with `ruby -w` you'll get a warning when you define a method multiple times. Doing an `undef_method` lets your code run `-w` clean. Attributing malice is unwarranted and unappreciated.

Please grow a sense of humor and/or a sense of humility. That's the only reason why `i_suck_and_my_tests_are_order_dependent!` exists. Nobody is being forced to use it and your belief that that is the case betrays your unwillingness to read the documentation.

### Workaround

Alternatively, instead of adding this patch to fix a problem that _almost nobody else has_ you could do one of two simple things:

1. Pin your version of minitest to 5.3.3. This will fix your problem entirely.
2. Actually fix your test bugs (tho [you claim not to have any](https://github.com/seattlerb/minitest/issues/635#issuecomment-237639109), so I'm not even sure why this PR exists). Anything that minitest is pointing out via randomization is a legit issue.

(possibly related... for the record, **minitest is a test framework**. It is *not* a documentation generation system. Never in my life have I seen a usable and valuable set of documentation generated from tests. It has never been a goal of minitest to try to meet that need nor will it ever be.)

For those folk who were hit by this change _two years ago_, they've resolved their issues with it and in multiple cases thanked me for pointing out extra bugs in their tests. This class shuffling issue has literally not come up in years.

> I have not added any test for this as there doesn't seem to be any place where this part of the test runner is currently being tested.

Not true. There are multiple tests for `test_order`. There are multiple tests for `run`. There are tests for (nearly?) all of this stuff. Please add tests when you submit a PR to a well tested project.

### Posted for transparency's sake

I got a message from @chrisnicola via github's email notification on an edited comment:

> There is no editorializing, but thanks for editing my comment to free it up from any possible criticism of the code. We wouldn't want reasonable discourse to occur.

@tenderlove and I are the only ones with permissions to edit comments here. I assure you we have not. We encourage healthy debate. This is not healthy debate at this point. This is toxic.

That message appears to have been edited back out and I'm not sure if that was for good reasons or not, but I'm putting it up here to document what we as maintainers have to deal with.

zenspider:
Also see #550 for an example of healthy discourse on this very topic.

PR #640, opened 2016-08-08
Fix spelling continous -> continuous

fgo:


zenspider:
Ugh. Good catch. Thanks. Done.

PR #642, opened 2016-08-11
Accept floating point inconsistencies in bench_exp

quixoten:
On JRuby:

```
>> JRUBY_VERSION
=> "9.1.2.0"
>> (Math.log10(1_000_000) / Math.log10(10))
=> 5.999999999999999
```

On MRI:

```
>> RUBY_VERSION
=> "2.3.0"
>> (Math.log10(1_000_000) / Math.log10(10))
=> 6.0
```

This causes bench_exp to give different results depending on the ruby implementation. I ran into it specifically when trying to benchmark differences between MRI and JRuby. I was making some false conclusions before I realized JRuby wasn't running in the same race.

### DIFF
--- a/lib/minitest/benchmark.rb
+++ b/lib/minitest/benchmark.rb
@@ -33,8 +33,8 @@ module Minitest
     #   bench_exp(2, 16, 2) # => [2, 4, 8, 16]
 
     def self.bench_exp min, max, base = 10
-      min = (Math.log10(min) / Math.log10(base)).to_i
-      max = (Math.log10(max) / Math.log10(base)).to_i
+      min = (Math.log10(min) / Math.log10(base)).ceil
+      max = (Math.log10(max) / Math.log10(base)).ceil
 
       (min..max).map { |m| base ** m }.to_a
     end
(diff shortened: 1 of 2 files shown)


zenspider:
Not sure I understand. Is the problem the _value_ actually calculated, or the _representation_ of the value?

quixoten:
Not sure how to best answer. If I recall correctly, in Jruby, `bench_exp(1, 1_000_000)` was giving me six items: `[1, 10, 100, 1_000, 10_000, 100_000]`, but on Ruby it returned seven items: `[1, 10, 100, 1_000, 10_000, 100_000, 1_000_000]`. My expectation was that `bench_exp(1, 1_000_000)` would return the same set of numbers on both Ruby and JRuby.  Maybe I just made a bad assumption?

Also, I should let you know that minitest is a pleasure to work with. It has been my goto testing framework for a long time. I write better tests when I use it, and that has improved my code as well. Thanks for letting me use it!

zenspider:
@headius @enebo ? any clue? I expected this just to be representation differences

enebo:
@zenspider https://github.com/jruby/jruby/issues/4089 .  As for representation log10 is a special method which corrects and gives us what we expect (but we are doing it wrong).  Check it out:
```text
mri23 -e 'p Math.log(1000000, 10)'
5.999999999999999
```

So we behave like generic log method in MRI and not the corrected log10 method (Java also happens to have an API to correct for human-friendly expected value).  This will be in our 9.1.3.0 release due out in next week.

zenspider:
This should be fixed on the jruby side.

PR #645, opened 2016-09-15
Expose options

alkalinecoffee:
This PR exposes the minitest `options` var.

Our team leverages minitest with a custom library to make live HTTP requests against our APIs. Particularly, we want access to the generated `seed` value so we can easily identify specific test runs in logs/splunk (our API test library would pass this value along in each request during the test run).

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -40,6 +40,11 @@ module Minitest
   mc.send :attr_accessor, :extensions
 
   ##
+  # The options used for this test run.
+
+  mc.send :attr_accessor, :options
+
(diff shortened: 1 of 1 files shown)


aiomaster:
LOL, just now I see this pull request that does very much the same like mine:
https://github.com/seattlerb/minitest/pull/657

Sorry @alkalinecoffee , that I had not seen this before.
It would be very cool if one of these two pull requests could get merged.

@zenspider: Should I close mine?
Only difference is, where exactly the options variable is set.

zenspider:

PR #646, opened 2016-09-15
Fixed code example with inproper casing

rvictory:
MiniTest was spelled as "Minitest" in the Readme example.


zenspider:
Sorry, but it is correct: https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L9

zenspider:
Sorry: https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L9

PR #648, opened 2016-09-22
Fix autorunning when $! is reset by an at_exit handler

russelldavis:
Normally, when using Minitest.autorun, the autorunning will be skipped if Ruby exits with an exception. However, if an at_exit handler (that runs before minitest's handler) raises and rescues an exception, it will reset `$!` back to nil, causing minitest to no longer skip autorunning. This PR fixes that.

Here's a quick way to reproduce:

```ruby
require 'minitest/autorun'

class TestThing < Minitest::Test
  def test_thing
    puts 'hi from test_thing'
  end
end

at_exit do
  raise 'bar' rescue nil
end

raise 'foo'
```

Without this PR, minitest will run the tests. With this PR, it will not.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -49,8 +49,11 @@ module Minitest
   # Registers Minitest to run at process exit
 
   def self.autorun
+    return if @@installed_at_exit
+
+    patch_at_exit
     at_exit {
-      next if $! and not ($!.kind_of? SystemExit and $!.success?)
+      next if @@exiting_with_error
 
       exit_code = nil
 
@@ -60,7 +63,7 @@ module Minitest
       }
 
       exit_code = Minitest.run ARGV
-    } unless @@installed_at_exit
+    }
     @@installed_at_exit = true
   end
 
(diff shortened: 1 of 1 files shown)


zenspider:
I'm really not comfortable with this patch. This scenario seems rare to me. Does it occur a lot for you? If so, how?

russelldavis:
Yeah, it's enough of a problem at Stripe that we've patched this internally.

It's actually pretty easy for it to happen. There are all sorts of reasonable things you could do in an at_exit handler that would cause it. It can even happen when your `at_exit` block doesn't have a `rescue` directly in it (for example, calling `FileUtils.rm_f('file_that_might_not_exist')`, which internally does it's own `rescue`). We can't exp

Another issue this fixes that I failed to mention earlier -- if you run the example script, you'll notice that it exits with a status code of 0. Even though loading the file failed with an exception. (This happens because minitest calls `exit(false)` when the tests finish successfully.) This is really bad -- it causes our CI server to report that everything worked fine, when in fact something very bad happened at load time.

I'm with you on not loving the patch (I try to avoid monkeypatching if at all possible), but I can't think of a better way to fix this (so in this case, I think it's the lesser of two evils).

Thoughts? (Thanks for taking a look.)

zenspider:
@russelldavis can you explain to me what your `at_exit`'s are for? Maybe we're going about this the wrong way?

russelldavis:
I'd rather avoid a back and forth about the details of my specific `at_exit`s; I don't think they're particularly relevant (and I assure you they're doing reasonable things).

To me, the core of the issue is that anyone could write something like `at_exit {FileUtils.rm_f('some_file')}`, and that nobody would reasonably catch during code review that, months down the line when `some_file` happens to not exist, it could end up masking other errors in CI in a way that's also very tricky to track down.

I think the case for fixing this is pretty strong, but I'll let you take it from here.

zenspider:
> I'd rather avoid a back and forth about the details of my specific at_exits; I don't think they're particularly relevant (and I assure you they're doing reasonable things).

...and I'd rather not. It's important to be able to know if this is an `at_exit` that is doing something to try to clean up the tests or if it is a general library `at_exit` that has nothing to do with the tests themselves. It could be that you've architected something incorrectly and it has a simple workaround. It could be that those `at_exit`s don't need to load at all for a test run. Without information, I'm shooting in the dark, and as it stands, this monkeypatch is too risky to warrant something that is edge case for 99+% of minitest usage.

russelldavis:
I disagree, but I'm done here.

zenspider:
What the fuck? You allergic to dialog?

PR #650, opened 2016-09-24
Write strings directly to the IO object

### LINKED ISSUE #649: incompatible character encodings: UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)
### Desc

Hi. It's very very rare case. Error appears in `Minitest::Test#to_s` or in `Minitest::SummaryReporter#aggregated_results` methods. 

Cause of the problem is a exception message returned from WIN32OLE object, if message contain not ASCII chars.

For look at this error should to run test case:

``` ruby
class EncodingCompatibilityErrorTest < Minitest::Test
  def test_this_is_not_ascii_assertion
    assert_equal "ЁЁЁ", "ёёё"
  end

  def test_this_is_non_ascii_failure_message
     fail 'ЁЁЁ'.force_encoding('ASCII-8BIT')
  end
end
```

fail:

```
/home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:597:in `join': incompatible character encodings: ASCII-8BIT and UTF-8 (Encoding::CompatibilityError)
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:597:in `aggregated_results'
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:582:in `report'
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:658:in `each'
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:658:in `report'
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:140:in `run'
        from /home/vlv/.rvm/gems/ruby-2.0.0-p645/gems/minitest-5.9.0/lib/minitest.rb:62:in `block in autorun'

```

Failure in the `Minitest::Test#to_s` appear when `self.location` contain non ASCII chars.
### Fix

For fix bug should force encoding exception message. I suppose `Minitest::Test#to_s` is good place for it:

``` ruby
    def to_s # :nodoc:
      return location if passed? and not skipped?

      failures.map { |failure|
#        "#{failure.result_label}:\n#{self.location}:\n#{failure.message}\n"
        "#{failure.result_label}:\n#{self.location}:\n#{failure.message\
           .force_encoding('utf-8')}\n"
      }.join "\n"
    end

```

tenderlove:
$stdout has binary output by default, so we can write any old bytes we'd
like to write.  Writing one test failure result at a time allows us to
avoid dealing with encoding issues when doing string concatenation

Fixes #649

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -593,7 +593,7 @@ module Minitest
       filtered_results = results.dup
       filtered_results.reject!(&:skipped?) unless options[:verbose]
 
-      filtered_results.each_with_index.each { |result, i|
+      filtered_results.each_with_index { |result, i|
         io.puts "\n%3d) %s" % [i+1, result]
       }
       io.puts
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -2,6 +2,7 @@ require "optparse"
 require "thread"
 require "mutex_m"
 require "minitest/parallel"
+require "stringio"
 
 ##
 # :include: README.rdoc
@@ -579,7 +580,7 @@ module Minitest
       io.puts unless options[:verbose] # finish the dots
       io.puts
       io.puts statistics
-      io.puts aggregated_results
+      aggregated_results io
       io.puts summary
     end
 
@@ -588,21 +589,20 @@ module Minitest
         [total_time, count / total_time, assertions / total_time]
     end
 
-    def aggregated_results # :nodoc:
+    def aggregated_results io # :nodoc:
       filtered_results = results.dup
       filtered_results.reject!(&:skipped?) unless options[:verbose]
 
-      s = filtered_results.each_with_index.map { |result, i|
-        "\n%3d) %s" % [i+1, result]
-      }.join("\n") + "\n"
-
-      s.force_encoding(io.external_encoding) if
-        ENCS and io.external_encoding and s.encoding != io.external_encoding
-
-      s
+      filtered_results.each_with_index.each { |result, i|
+        io.puts "\n%3d) %s" % [i+1, result]
+      }
+      io.puts
+      io
     end
 
-    alias to_s aggregated_results
+    def to_s
+      aggregated_results(StringIO.new(''.b)).string
+    end
 
     def summary # :nodoc:
       extra = ""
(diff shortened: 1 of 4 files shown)


lib/minitest.rb
```diff
@@ -588,21 +589,20 @@ def statistics # :nodoc:
...
-      }.join("\n") + "\n"
-
-      s.force_encoding(io.external_encoding) if
-        ENCS and io.external_encoding and s.encoding != io.external_encoding
-
-      s
+      filtered_results.each_with_index.each { |result, i|
+        io.puts "\n%3d) %s" % [i+1, result]
+      }
+      io.puts
```
nateberkopec: Nice cleanup of the original code too 👍

test/minitest/metametameta.rb
```diff
@@ -14,7 +14,7 @@ class MetaMetaMetaTestCase < Minitest::Test
   def run_tu_with_fresh_reporter flags = %w[--seed 42]
     options = Minitest.process_args flags
 
-    @output = StringIO.new("")
+    @output = StringIO.new("".encode('UTF-8'))
```
zenspider: This needs to be guarded too or it'll have to wait for MT6.

lib/minitest.rb
```diff
@@ -613,6 +613,14 @@ def summary # :nodoc:
       "%d runs, %d assertions, %d failures, %d errors, %d skips%s" %
         [count, assertions, failures, errors, skips, extra]
     end
+
+    private
+
+    if '<3'.respond_to? :b
```
zenspider: Also needs a guard.

tenderlove:
> I think I'd rather keep this code as-is and wait to merge for MT6. Otherwise, this needs to be protected w/ extra guards.

You mean guards for Ruby 1.8?  CI doesn't seem to be running it, and installing it is a huge pain.  Rather than claiming some code works on 1.8 without testing, I'd rather wait for MT 6. 😄

zenspider:
Done. Thanks!

PR #655, opened 2016-11-07
Add assert_boolean and refute_boolean

aycabta:
I can not write below code with MiniTest;

```ruby
assert_instance_of(Boolean, value)
```

because Ruby doesn't have `Boolean` class, `true` is the only instance of `TrueClass` and `false` is the only instance of `FalseClass`.

On another front, [test-unit has `assert_boolean`](https://github.com/test-unit/test-unit/blob/c31f9d017eaf899331fadf4c5bb4943bea7ed3c6/lib/test/unit/assertions.rb#L1193-L1207). I think that the assertion is so useful.

```ruby
assert_boolean(true) # => pass
assert_boolean(false) # => pass
assert_boolean(nil) # => fail
```

This Pull Request adds `assert_boolean` and `refute_boolean` to MiniTest.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -207,6 +207,16 @@ module Minitest
     end
 
     ##
+    # Fails unless +obj+ is true or false.
+
+    def assert_boolean obj, msg = nil
+      msg = message(msg) {
+        "Expected #{mu_pp(obj)} to be true or false"
+      }
+      assert [true, false].include?(obj), msg
+    end
+
+    ##
     # Fails unless +collection+ includes +obj+.
 
     def assert_includes collection, obj, msg = nil
@@ -572,6 +582,16 @@ module Minitest
 
     def refute_in_epsilon a, b, epsilon = 0.001, msg = nil
       refute_in_delta a, b, a * epsilon, msg
+    end
+
+    ##
+    # Fails if +obj+ is true or false.
+
+    def refute_boolean obj, msg = nil
+      msg = message(msg) {
+        "Expected #{mu_pp(obj)} to not be true or false"
+      }
+      refute [true, false].include?(obj), msg
     end
 
     ##
--- a/lib/minitest/expectations.rb
+++ b/lib/minitest/expectations.rb
@@ -57,6 +57,15 @@ module Minitest::Expectations
   infect_an_assertion :assert_in_epsilon, :must_be_within_epsilon
 
   ##
+  # See Minitest::Assertions#assert_boolean
+  #
+  #    collection.must_be_boolean
+  #
+  # :method: must_be_boolean
+
+  infect_an_assertion :assert_boolean, :must_be_boolean, :unary
+
+  ##
   # See Minitest::Assertions#assert_includes
   #
   #    collection.must_include obj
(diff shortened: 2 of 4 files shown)


aycabta:
Add `must_be_boolean` and `wont_be_boolean` too.

nateberkopec:
Just out of curiousity, where did you find it necessary to test for a boolean return type rather than just assert the truthiness or falsiness?

zenspider:
Yeah. I'm having trouble seeing a use case for this. Seems like you either want `assert`/`refute` for truthy/falsey or you want `assert_equal` to just say what the value should be `true`/`false`. It's not like you don't know, right?

nateberkopec:
Yeah, that's what I'm struggling to figure out. In what case would you *not* know it's true or false?

aycabta:
> Just out of curiousity, where did you find it necessary to test for a boolean return type rather than just assert the truthiness or falsiness?

For example, [my gem needs it](https://github.com/aycabta/deviantart/blob/a8b800bde5c2c014ea563d69f3c21a36bda5e468/test/user_test.rb#L101-L102) for [REST API](https://www.deviantart.com/developers/http/v1/20160316/user_statuses_list/16d6e7a9e6425b0024aa6ede9811452d). I can not know which I take beforehand.

> EDIT: I ask because, as Matz has said, [everything in Ruby is a boolean](https://www.ruby-forum.com/topic/4412411#1103364).

After this, Matz said [a more in-depth description about it in Japanese](http://qiita.com/yatemmma/items/3597a0e20556fb846d3f), and [it's translated into English](https://bugs.ruby-lang.org/issues/12515#note-7) by @shugo.

> At http://qiita.com/yatemmma/items/3597a0e20556fb846d3f, Matz explained the reason why Ruby doesn't have Boolean.
> - true and false are just representative samples of boolean values.
> - There is no method to be defined in Boolean (e.g., implementations of TrueClass#& and FalseClass#& are different).
> - The use case of Boolean is therefore limited to kind_of? checks, and it prevents duck typing.

Matz strongly recommends using duck typing for TrueClass / FalseClass like others. It's the reason of Ruby doesn't have Boolean.

I just propound assertions for true/false verification for a value what is not known beforehand on this Pull Request. It has no relation to a concept of TrueClass / FalseClass design by Matz.

zenspider:
Except you do know: https://github.com/aycabta/deviantart/blob/2adbdfe93136a0d7e8a8a43ae221d8bc068b86c2/test/fixtures/user_statuses.json#L2

aycabta:
The `test/fixtures/user_statuses.json` is just for mock test. You can run [real access test what `README.md` describes how to](https://github.com/aycabta/deviantart/blob/2adbdfe93136a0d7e8a8a43ae221d8bc068b86c2/README.md#how-to-test). The value `has_more` will be `true` after the statuses increase over the page size. There sometimes are the values like this; received data from network, ramdomize data, hardware data and others in API. The `assert_boolean` is useful in those cases.

zenspider:
Thanks, but I just don't think this fits in minitest.

You're welcome to publish this as a gem and add it to my README's list of extensions.

aycabta:
OK, thanks a lot. I [use `assert_includes` with `[true, false]` instead of that](https://github.com/aycabta/deviantart/blob/7c8068f62c2f5b368bb440ec1724a741394e542a/test/user_test.rb#L102).

PR #657, opened 2016-11-17
Make the testopts accessible in your tests.

aiomaster:
Hello,

I would like to have the options available to use in my tests.
Especially the seed value is very interesting for me.
I would like to make some tests use different, random but reproducable input everytime the tests run to cover more test cases.
The seed value is good to support that kind of stuff, but I have to access it somehow in my test cases.
By exposing the options as attr_accessor I can use it.

Thanks!

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -47,6 +47,10 @@ module Minitest
   self.info_signal = "INFO"
 
   ##
+  # Access the testopts for example to use the seed value in your tests.
+  mc.send :attr_accessor, :testopts
+
+  ##
   # Registers Minitest to run at process exit
 
   def self.autorun
@@ -121,6 +125,7 @@ module Minitest
     self.load_plugins unless args.delete("--no-plugins") || ENV["MT_NO_PLUGINS"]
 
     options = process_args args
+    self.testopts = options
 
     reporter = CompositeReporter.new
     reporter << SummaryReporter.new(options[:io], options)
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -47,8 +47,8 @@ module Minitest
   self.info_signal = "INFO"
 
   ##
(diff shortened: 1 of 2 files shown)


zenspider:
For my own understanding... You want to use `srand` or something with a random but reproducible value?

lib/minitest.rb
```diff
@@ -46,6 +46,10 @@ module Minitest
   self.info_signal = "INFO"
 
   ##
+  # Access the testopts for example to use the seed value in your tests.
+  mc.send :attr_accessor, :testopts
```
zenspider: "testopts" isn't a word, nor are they specific to testing. Please switch this to "option" (EDIT: singular).

aiomaster: Why singular and not "options"? There is more than one option in that hash.

aiomaster:
You are right, I basically want to use srand with this seed value to set a reproducable but random state for my tests.
In combination for example with the Faker gem (https://github.com/stympy/faker) which uses internally the random class of ruby to generate different types of data you can have more powerfull tests that don't test exactly the same input again and again, but more cases where you couldn't think of when creating your tests. But it is very important, that you can reproduce the values in case of a failure.

sandstrom:
friendly ping @aiomaster @zenspider  😄 

This would be very useful, I'd also like to access the seed value.

aiomaster:
I rebased from current master.

lib/minitest.rb
zenspider: You only access them one at a time.

andrelaszlo:
We're also using the Faker gem, and would like to seed it with the same seed as Minitest - there are some intermittent failures that are really annoying to track down otherwise.

This feature would save us a lot of time, @zenspider! Is there something you disagree with in this implementation? I'd be happy to fix it if @aiomaster isn't available.

zenspider:

PR #662, opened 2016-12-07
WIP: fix minitest/hell

azul:
Since Minitest::Test was never extended with
  Minitest::Parallel::Test::ClassMethods
changing the test_order had no effect.
Once you extend it a test_order of :random has no effect either.

This fix includes a test that confirms minitest/hell does what it is
supposed to do. However requiring this test will require minitest/hell
and thus break the other tests. I tried to work around this by setting
the test order explecitly in those tests but it does not seem to help.

### DIFF
--- a/lib/minitest/hell.rb
+++ b/lib/minitest/hell.rb
@@ -1,6 +1,9 @@
 require "minitest/parallel"
 
 class Minitest::Test
+  include Minitest::Parallel::Test
+  extend Minitest::Parallel::Test::ClassMethods
+
   class << self
     alias :old_test_order :test_order # :nodoc:
 
--- a/test/minitest/test_minitest_hell.rb
+++ b/test/minitest/test_minitest_hell.rb
@@ -1,0 +1,88 @@
+require 'minitest/metametameta'
+require 'minitest/hell'
+
+class TestMinitestHell < MetaMetaMetaTestCase
+  def self.test_order
+    :random
+  end
+
+  require "monitor"
+
+  class Latch
+    def initialize count = 1
+      @count = count
+      @lock  = Monitor.new
+      @cv    = @lock.new_cond
+    end
+
+    def release
+      @lock.synchronize do
+        @count -= 1 if @count > 0
+        @cv.broadcast if @count == 0
+      end
+    end
+
+    def await
+      @lock.synchronize { @cv.wait_while { @count > 0 } }
+    end
+  end
+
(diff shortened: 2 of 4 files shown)


lib/minitest/hell.rb
```diff
@@ -1,6 +1,9 @@
 require "minitest/parallel"
 
 class Minitest::Test
+  include Minitest::Parallel::Test
+  extend Minitest::Parallel::Test::ClassMethods
+
```
zenspider: I addressed this by removing the whole class body and declaring `parallelize_me!` instead. That way there's only one implementation anywhere.

test/minitest/test_minitest_test.rb
```diff
@@ -146,6 +146,9 @@ def test_inherited_hook_plays_nice_with_others
 
 class TestMinitestRunner < MetaMetaMetaTestCase
   # do not parallelize this suite... it just can't handle it.
+  def self.test_order
+    :random
+  end
```
zenspider: I don't like these... Let me think about an alternative... but I'm thinking of something that runs `test_minitest_hell` entirely on its own.

test/minitest/test_minitest_hell.rb
```diff
@@ -0,0 +1,88 @@
+require 'minitest/metametameta'
+require 'minitest/hell'
+
+class TestMinitestHell < MetaMetaMetaTestCase
+  def self.test_order
+    :random
+  end
```
zenspider: What's the point of this? There's only 1 test in here anyways...

test/minitest/test_minitest_test.rb
azul: Yes. That's probably the way to go.

test/minitest/test_minitest_hell.rb
azul: The test is basically a copy of the parallelize me test from https://github.com/seattlerb/minitest/blob/master/test/minitest/test_minitest_test.rb#L544. Since that was in a test case that could not be parallelized i copied that property as well just in case more tests get added.

Frankly i do not quite understand how this test works and was mostly trying to come up with something working from what existed before.

zenspider:
I think this has been addressed... yes? Please close if so.

azul:
Yes, The issue itself is addressed. Testing is most likely painful due to the nature of minitest/hell. I won't get to work on it so i'll close this.

PR #672, opened 2017-01-01
Expand MT6 to Minitest 6.

xaviershay:
It's non-obvious what the acronym means, particularly when minitest has
been pulled in via a different library (such as Rails) and the user may
not be aware they are using minitest.

This confused at least me and one other person (on
https://github.com/rspec/rspec-rails/issues/1761)


zenspider:
Fair enough. Done. Thanks!

P.S. I also just committed a change to use warn for that line so `-W0` will silence them.

PR #675, opened 2017-01-05
Workaround for rdoc bug in test.rb

Paxa:
Fixes https://github.com/seattlerb/minitest/issues/674


zenspider:
UGH. Thank you. Hopefully that'll get fixed in rdoc soon, but I won't hold my breath.

I should go check my other projects to see if they suffer this problem.

PR #676, opened 2017-01-17
README.rdoc updates for common Windows problems

snarfmason:
I've made a section under FAQ covering the two main points I ran into when trying to run Minitest's tests on Windows. It occurred to me that particularly the encoding one is not entirely Windows-specific, so I just called the section Developing Minitest.


zenspider:
Sorry for taking so long... I finally got this in. Thanks!

PR #687, opened 2017-05-02
Fixes to be ok with minitest-line plugin

shemusok:
1. I've found, that minitest-line plugin is not OK with minitest, if launched through spring binstub as:
`./bin/rake test test/some_test.rb -l 42`
because plugin can't recognize test file name.
So this fixed with additional option :file 124 line.

2. I've found, that test with '/' in 'describe' part is not OK with regexp in 297 line,
so I've changed it to match only if slashes are first and last symbols.

3. During testing I've found that it's impossible to use minitest right from github, so I've created gemspec file.

So, now launching single test given by line number is OK (with corresponding [fixes in minitest-line plugin](https://github.com/judofyr/minitest-line/pull/14)) with three popular launching commands:
`./bin/rake test test/some_test.rb -l 42`
`rake test test/some_test.rb -l 42`
`ruby -Itest test/some_test.rb -l 42`

@zenspider review pls! :)

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -121,6 +121,9 @@ module Minitest
     self.load_plugins
 
     options = process_args args
+    if file = args.find {|arg| File.file?(arg)}
+      options.merge!(file: file)
+    end
 
     reporter = CompositeReporter.new
     reporter << SummaryReporter.new(options[:io], options)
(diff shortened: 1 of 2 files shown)


zenspider:
No response from OP. Closing.

PR #693, opened 2017-05-15
Mock#expect: Add documentation about multiple calls to same method

Insti:
Add documentation to `Minitest::Mock#expect` about how to test multiple calls to the same method.


zenspider:
Fantastic. Thank you!

PR #696, opened 2017-05-19
Handle empty backtrace from external filter

iNecas:
In the wild the backtrace_filter can be set by external filter (Rails do
it for their silencer). This can lead to the case there the external
backtrace_filer returns empty array. Other parts of minitest count on
the fact that there will still be something in the backtrace to use, and
can lead to issues such as "undefined method `split' for nil:NilClass"
when producing deprecation warnings.

Since we already had the logic to use all the backtrace when filtering
too much, I've moved this logic to be used also for external filters.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -231,7 +231,9 @@ module Minitest
   end
 
   def self.filter_backtrace bt # :nodoc:
-    backtrace_filter.filter bt
+    new_bt = backtrace_filter.filter bt
+    new_bt = bt.dup if new_bt.empty?
+    new_bt
   end
 
   ##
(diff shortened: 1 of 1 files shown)


k0kubun:
By https://github.com/rails/rails/pull/29572 and upgrading Rails from 5.1.2 -> 5.1.3, I hit an error (`undefined method 'split' for nil:NilClass`) by `Rails.backtrace_cleaner`. I hope this is merged.

jacobo:
seeing this as seemingly just a side effect of Rails (5.2.3) being loaded. No application code setting a custom backtrace filter or anything like that. Checking the value of `Minitest.backtrace_filter` at runtime shows a `Rails::BacktraceCleaner` object

Issue #631, opened 2016-07-08
Clarification on StatisticsReporter's Public API


sshaw:
[The docs for `StatisticsReporter`](http://docs.seattlerb.org/minitest/Minitest/StatisticsReporter.html) say:

> ... If you want to create an entirely different type of output (eg, CI, HTML, etc), this is the place to start.

This indicates that it is part of the public API but the accessors one needs to use in order to create an entirely different type of output [have `stopdoc` directives](https://github.com/seattlerb/minitest/blob/4e146b1515b19a5c474e39c174b037510f6dbc6e/lib/minitest.rb#L497-L506).  Is this intentional or are these accessors safe to use in subclasses?

zenspider:
Seems counterintuitive, doesn't it? Looking...

zenspider:
Nope. Makes no sense. I was probably being lazy. Consider them public and feel free to throw me a PR with doco on them.

sshaw:
>  Consider them public and feel free to throw me a PR with doco on them.

Will do.

zenspider:
This seems hondled

PR #704, opened 2017-08-26
Doc updates: reporter usage, mention Hoe for dev

sshaw:
Stems from #631. 

The current `StatisticsReporter`  paradigm requires subclasses to call super to get the aggregated failures, etc... May be better to add subclass hook E.g., `output_results`.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -422,7 +422,8 @@ module Minitest
 
     ##
     # Returns a single character string to print based on the result
-    # of the run. Eg ".", "F", or "E".
+    # of the run. One of <tt>"."</tt>, <tt>"F"</tt>,
+    # <tt>"E"</tt> or <tt>"S"</tt>.
 
     def result_code
       raise NotImplementedError, "subclass responsibility"
@@ -560,8 +561,10 @@ module Minitest
     end
 
     ##
-    # Record a result and output the Runnable#result_code. Stores the
-    # result of the run if the run did not pass.
+    # Output and record the result of the test. Call
+    # {result#result_code}[rdoc-ref:Runnable#result_code] to get the
+    # result character string. Stores the result of the run if the run
+    # did not pass.
 
     def record result
     end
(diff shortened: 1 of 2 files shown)


lib/minitest.rb
```diff
@@ -507,9 +510,21 @@ def record result # :nodoc:
...
+  #
+  #       print "<testsuite "
+  #       print "tests='#{count}' "
+  #       print "failures='#{failures}' "
+  #       # Remaining XML...
+  #     end
+  #   end
 
   class StatisticsReporter < Reporter
-    # :stopdoc:
```
zenspider: This drops `rake dcov` from 100% to 96.24%. ☹️

lib/minitest.rb
```diff
@@ -548,7 +562,8 @@ def record result # :nodoc:
       results << result if not result.passed? or result.skipped?
     end
 
-    def report # :nodoc:
+    # Report on the tracked statistics.
```
zenspider: `##` before and a blank line after, please.

lib/minitest.rb
```diff
@@ -439,8 +440,10 @@ def prerecord klass, name
     end
 
     ##
-    # Record a result and output the Runnable#result_code. Stores the
-    # result of the run if the run did not pass.
+    # Output and record the result of the test. Call
+    # {result#result_code}[rdoc-ref:Runnable#result_code] to get the
```
zenspider: This should link cleanly with just `+result_code+`, no?

sshaw: That just formats it as code.

lib/minitest.rb
sshaw: I can add docs to the `attr_accessor`s, but they're already self explanatory, don't you think?

zenspider:
No. Either they're excluded from the doco, or they're documented. `rake dcov` must be 100% to publish.

sshaw:
Okie dokie. Updated.

lib/minitest.rb
zenspider: Huh... that used to work in rdoc...

lib/minitest.rb
zenspider: no. I don't think. As you can see in the rest of the library.

sshaw: Then as someone that does (from time to time) I would suggest ignoring the coverage here.

lib/minitest.rb
sshaw: Checked again in RDoc 4.3.0 and 6.0.1; no links. The funny thing is Darkfish formats `+result_code+` with `code` tags and gives it this styling: 

```css
main .method-detail {
    margin-bottom: 2.5em;
    cursor: pointer;
}
```

So it still can look appear to be a link but, it's not. 😿

zenspider:
Done! Thank you!

PR #714, opened 2017-10-20
Docfix typo: Widnows

rickhull:
Including a second commit which removes trailing whitespace.  My editor does this automatically.


zenspider:
I split this in two so the whitespace changes are their own. Thanks!

PR #715, opened 2017-10-20
add bench_performance_{logarithmic,power} for spec-style benchmarks

rickhull:
- bring to parity with assert style benchmarks
- these wrappers have just gotten overlooked until now


zenspider:
Done. Thanks!

PR #716, opened 2017-10-24
Spec marshal

tenderlove:
This is a test case showing that instances of spec style tests can't be
marshaled due to their anonymous class.  This is problematic for
distributed tests as test information is stored on the instance of the
test object.

Our current solution at work is to copy all the information from the test in to a value object that we send across the wire.  Another problem is that the minitest reporters that are on the other side of the wire have certain expectations about the object passed to them: namely that the object is actually an instance of the test.  The reporter uses the class of the "test instance" for reporting, so we have to do weird stuff to fake out the reporter.

I've added a possible fix in 1e353c63b147907424a12cb0ca46ce5269721990 that is similar to what we're doing in our test suite.  It feels like the default implementation of `run` shouldn't return the test instance, but a concrete read-only value object that contains test run information instead.  Then we could eliminate the `marshal_dump` / load methods on the Test class.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -98,10 +98,54 @@ class Minitest::Spec < Minitest::Test
     Thread.current[:current_spec]
   end
 
+  # :nodoc:
+  #
+  # Get around restriction in `run_one_method` so that the marshalable instance
+  # will work.
+  def self.=== other
+    super(other) || (name == other.class.name && desc == other.class.desc)
+  end
+
   def initialize name # :nodoc:
     super
     Thread.current[:current_spec] = self
   end
+
+  def run
+    result = super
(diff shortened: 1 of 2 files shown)


zenspider:
Possible counter: https://gist.github.com/zenspider/4f1b0c9fbad6b718b60ddf8d4a5589e1

This removes custom marshal_load/dump entirely in favor of a Runnable called Result that pulls over the plain data that should always be clean to marshal. It's bigger than yours because it cleans up a fair amount of test output as well.

tenderlove:
Oh hello.  I'd forgotten about this, but we're running in to it at work now.  You patch seems fine to me, but I think it's weird that `Result` inherits from `Runnable` because I don't think a Result is really Runnable, is it?

zenspider:
I've got this working with a slightly different solution.

PR #717, opened 2017-10-25
fix comment: [1, 2, 3].sum is 6, not 7

rickhull:
WTB easter egg prize


zenspider:
IIRC I put that in there intentionally to see if anyone would catch it... 7 years and 11 days later...

joelparkerhenderson:
Great egg! :)

zenspider:
Done. Thanks!

PR #718, opened 2017-10-31
add benchmark tests

rickhull:
- test_sigma
- test_fit_error
- test_fit_linear_x2 (fails, rr for fit_linear(x2) is over 0.99)
- test_fit_linear_x3 (fails, rr for fit_linear(x3) is over 0.99)

Per discussion on IRC.  Some background:

I have been playing with minitest's fit methods in https://github.com/rickhull/compsci/blob/master/lib/compsci/fit.rb

These tests in this PR originate from that effort.  `test_sigma` and `test_fit_error` because why not, and also per #717 where the comments for `Benchmark#sigma` show an incorrect example.

Separately, I discovered that the r^2 (`rr` or `r2`) value from `fit_linear` is not that useful for distinguishing linear data from x^2 or even x^3.  `test_fit_linear_x2` and `test_fit_linear_x3` fail when I think we really want them to succeed.  See #719 for more details.

### DIFF
--- a/test/minitest/test_minitest_benchmark.rb
+++ b/test/minitest/test_minitest_benchmark.rb
@@ -118,6 +118,32 @@ class TestMinitestBenchmark < Minitest::Test
     assert_fit :power, x, y, 0.96, 3.119e-5, 0.8959
   end
 
+  def test_sigma
+    bench = Minitest::Benchmark.new :blah
+    assert_equal bench.sigma([1,2,3]), 6
+    assert_equal bench.sigma([1,2,3]) { |n| n**2 }, 14
+  end
+
+  def test_fit_error
+    bench = Minitest::Benchmark.new :blah
+    assert_equal bench.fit_error([[1, 1], [2, 2], [3, 3]]) { |x| x }, 1.0
+    assert_in_delta bench.fit_error([[1, 1], [2, 2], [3, 4]]) { |x| x }, 0.785
+  end
+
+  def test_fit_linear_x2
+    bench = Minitest::Benchmark.new :blah
+    xs = [1, 10, 100, 1000]
+    _a, _b, rr = bench.fit_linear(xs, xs.map { |x| x**2 }), :<, 0.99
+    assert_operator rr, :<, 0.99
+  end
+
+  def test_fit_linear_x3
+    bench = Minitest::Benchmark.new :blah
+    xs = [1, 10, 100, 1000]
+    _a, _b, rr = bench.fit_linear(xs, xs.map { |x| x**3 })
+    assert_operator rr, :<, 0.99
+  end
+
   def assert_fit msg, x, y, fit, exp_a, exp_b
     bench = Minitest::Benchmark.new :blah
 
--- a/test/minitest/test_minitest_benchmark.rb
+++ b/test/minitest/test_minitest_benchmark.rb
@@ -144,6 +144,28 @@ class TestMinitestBenchmark < Minitest::Test
     assert_operator rr, :<, 0.99
   end
 
+  def test_fit_constant_beyond_linear
+    bench = Minitest::Benchmark.new :blah
+    # This is actual data from a Minitest::Benchmark run (testing Tree#push)
+    # It sure looks steeper than linear slope 1.
+    # Look at y go from .05 to .26 as x goes from 1000 to 2000
+    xs = [10, 100, 1000, 2000]
+    ys = [0.000092, 0.000618, 0.050752, 0.258891]
+
+    # I don't know how to call #assert_performance_constant on static data
+    # i.e. without doing timing
+    # Here is what assert_performance_constant tests
+    _a, b, _rr = bench.fit_linear(xs, ys)
+    assert_in_delta 0, b, 0.01
+
+    # The assertion below is intended to be a synthetic version of
+    # #refute_performace_constant
+    refute_in_delta 0, b, 0.01
+
+    # Interestingly, assert_performance_constant succeeds for this data
+    # while assert_performance_linear fails.
+  end
+
   def assert_fit msg, x, y, fit, exp_a, exp_b
     bench = Minitest::Benchmark.new :blah
 


rickhull:
added a new test `test_fit_constant_beyond_linear` that is expected to fail under the current regime, as with `test_fit_linear_x2` and `test_fit_linear_x3`

In this case, I have static xs and ys that were gathered from actual testing with Minitest::Benchmark.  The ys are beyond linear with an increasing slope well beyond 1.  Yet the test for assert_performance_constant succeeds.  We would want assert_performance_constant to fail for this data.

rickhull:
Here are the failures, as expected:

```
Finished in 0.097131s, 3263.6341 runs/s, 9430.5641 assertions/s.

  1) Error:
TestMinitestBenchmark#test_fit_constant_beyond_linear:
NameError: undefined local variable or method `b' for #<TestMinitestBenchmark:0x007fa183568340>
    /home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:159:in `test_fit_constant_beyond_linear'

  2) Failure:
TestMinitestBenchmark#test_fit_linear_x2 [/home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:137]:
Expected 0.99 to be < 0.99.

  3) Failure:
TestMinitestBenchmark#test_fit_linear_x3 [/home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:144]:
Expected 0.9916277794417635 to be < 0.99.
```

rickhull:
Here are the failures, as expected:

```
  1) Failure:
TestMinitestBenchmark#test_fit_linear_x3 [/home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:144]:
Expected 0.9916277794417635 to be < 0.99.

  2) Failure:
TestMinitestBenchmark#test_fit_constant_beyond_linear [/home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:163]:
Expected |0 - 0.0001262684384132403| (0.0001262684384132403) to not be <= 0.01.

  3) Failure:
TestMinitestBenchmark#test_fit_linear_x2 [/home/ubuntu/minitest/test/minitest/test_minitest_benchmark.rb:137]:
Expected 0.99 to be < 0.99.
```

rickhull:
Er, despite my previous comments and the comments in the test for `test_fit_constant_beyond_linear` -- I guess the slope is not over 1 anywhere.  At the highest slope, xdiff is 1000 and ydiff is 0.2 or whatever.  So what about the fact that when x doubles (from 1000 to 2000), y quintuples?  We know that's not constant behavior, and it shouldn't be linear either, right?  Sorry, my thinking isn't quite clear on this yet, but I'm pretty sure something isn't quite right.

rickhull:
Another angle on this: if we assert linear performance, I think what we really mean is that the slope is not increasing.  We are mostly saying it's "at worst linear".  I don't think we care so much if the slope is 0, 0.01, or 10.0.  The problem with r^2 for a linear fit is that it's designed to distinguish from constant behavior, not from increasing slope behavior.

If this is a fruitful line of thought, then maybe we want to look at the slopes between data points, like the derivative of our data, at least in the `assert_performance_linear` case.

test/minitest/test_minitest_benchmark.rb
```diff
@@ -118,6 +118,54 @@ def test_fit_power_noisy
...
+  def test_fit_error
+    bench = Minitest::Benchmark.new :blah
+    assert_equal bench.fit_error([[1, 1], [2, 2], [3, 3]]) { |x| x }, 1.0
+    assert_in_delta bench.fit_error([[1, 1], [2, 2], [3, 4]]) { |x| x }, 0.785
+  end
+
+  def test_fit_linear_x2
+    bench = Minitest::Benchmark.new :blah
+    xs = [1, 10, 100, 1000]
+    _a, _b, rr = bench.fit_linear(xs, xs.map { |x| x**2 }), :<, 0.99
```
zenspider: I dunno what this line is supposed to be doing, but if you remove "`, :<, 0.99`" it'll at least provide the real `rr`. As it currently stands, `_a` is populated with the 3 values from `fit_linear`, `_b` gets `:<`, and `rr` gets `0.99`, which means the test always fails.

zenspider: Both of these `fit_linear` tests (x2 and x3) pass if given a linear range... You're giving it an exponential range but expecting a linear response in `y`. In this test, if you get the real `rr` back, for a range of `1..10`, the `rr` is `0.94` which is what wolfram alpha says it should be:

http://www.wolframalpha.com/input/?i=linear+fit+1,+4,+9,+16,+25,+36,+49,+64,+81,+100

(that's so cool!)

Where with your exponential `xs`, mathematica (I couldn't figure out how to specify custom x values) agrees with minitest's argorithm that it fits:

```
In[28]:= LinearModelFit[{{1, 1}, {10, 100}, {100, 10000}, {1000, 
    1000000}}, x, x]["RSquared"]

Out[28]= 0.993096

In[29]:= LinearModelFit[{{1, 1}, {2, 100}, {3, 10000}, {4, 1000000}}, 
  x, x]["RSquared"]

Out[29]= 0.607999
```

test/minitest/test_minitest_benchmark.rb
```diff
@@ -118,6 +118,54 @@ def test_fit_power_noisy
...
+    # It sure looks steeper than linear slope 1.
+    # Look at y go from .05 to .26 as x goes from 1000 to 2000
+    xs = [10, 100, 1000, 2000]
+    ys = [0.000092, 0.000618, 0.050752, 0.258891]
+
+    # I don't know how to call #assert_performance_constant on static data
+    # i.e. without doing timing
+    # Here is what assert_performance_constant tests
+    _a, b, _rr = bench.fit_linear(xs, ys)
+    assert_in_delta 0, b, 0.01
```
zenspider:
```
In[34]:= LinearModelFit[{{10, 9.2*10^-5}, {100, 0.000618}, {1000, 
    0.050752}, {2000, 0.258891}}, x, x]["RSquared"]

Out[34]= 0.907874
```

which is what minitest is outputting.

zenspider: Here's what it looks like:

screen shot 2017-12-09 at 20 43 11

I'm not sure what this test is trying to say... it is indeed NOT linear and NOT constant... But the `xs` is very oddly picked and, if anything, too few are available to get a good fit that is free from runtime / machine noise.

zenspider:
screen shot 2017-12-09 at 21 05 08

rickhull:
We had a long discussion on #ruby IRC.  I think we're on the same page now -- that my tests are showing what I want to show, and that your math is correct.  I think we agree that my tests show that the linear r^2 is not that useful for distinguishing O(N) behavior from O(N^2) or even O(N^3).  I still think the best way to use the linear r^2 for asserting O(N) behavior is to make sure that exponential r^2 and power r^2 aren't higher.

Is there anything else you want from me or this PR?  I am expecting this to be closed without merging.  I can submit a new PR with just the successful `test_fit_error` and `test_fit_sigma` additions.

zenspider:
It's been way too long since this discussion happened. I agree with your last comment that we're both on the same page and that the math is correct but isn't too useful for linear fits. I suspect this comes down to the original issue's declaration of "I am not a statistician" (nor am I)... someone with more of a stats background could probably weigh in better than either of us.

Given the time, I'm gonna close.

I'm open to another PR improving the situation... but I don't know what that would be at this point. I also understand if, after 5 freakin' years, you might not be interested. Sorry about that.

rickhull:
No worries. I incorporated this idea into `Fit.best` here https://github.com/rickhull/compsci/blob/master/lib/compsci/fit.rb#L22

`assert_linear` becomes something like: `_(Fit.best(get_data())).must_equal :linear`

zenspider:
That looks really nice!

PR #721, opened 2017-11-10
Add FAQ example for setup before all tests

joelparkerhenderson:
I searched for a way to accomplish this, and found some blog posts about using a class variable such as @@foo = begin...end. 

When I tried using a class variable, `rake spec` reports "warning: class variable access from toplevel". 

When I tried a constant, `rake spec` succeeds without any warnings.

This setup approach has sped up some of my tests by 80%, because of database seeding and also because of data that's fetched across the internet.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -434,6 +434,21 @@ you want to extend your test using setup/teardown via a module, just
 make sure you ALWAYS call super. before/after automatically call super
 for you, so make sure you don't do it twice.
 
+=== How to run code before a group of tests?
+
+Use a constant with begin...end like this:
+
+  describe Blah do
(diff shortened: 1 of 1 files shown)


zenspider:
Ah. I was curious why you said you got warnings because of class vars, but then it makes sense given your patch uses minitest/spec. You could use a class-ivar instead w/ an accessor too.

joelparkerhenderson:
Thanks! Perhaps it could be good to add a similar test-syntax example that shows `@@foo`, and/or a spec-syntax example that shows an accessor `def self.foo`. Those ways both sound preferable for the general case.

zenspider:
Done. Thanks!

PR #727, opened 2017-12-23
For recent Ruby, show nested errors properly in backtraces

noahgibbs:
Recent Ruby versions store exceptions as Exception#cause when a different error is re-raised from a rescue block. This allows tracing a chain of exceptions to find out what inner exception is being "wrapped."

That's very useful! It's even more useful if Minitest actually displays the nested backtrace in a useful way. This pull request shows one way to make that happen.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -836,8 +836,24 @@ module Minitest
     end
 
     def message # :nodoc:
-      bt = Minitest.filter_backtrace(self.backtrace).join "\n    "
-      "#{self.exception.class}: #{self.exception.message}\n    #{bt}"
+      if self.exception.respond_to?(:cause) && self.exception.cause
+        # Build a chain of exception causes starting from the outermost
+        exc = self.exception
+        cause_chain = []
+        while exc
+          cause_chain.push(exc)
+          exc = exc.cause
+        end
+
+        bt_lines = cause_chain.map { |c|
+          [c.message] + Minitest.filter_backtrace(c.backtrace)
+        }.inject() { |acc, bt| acc + ["...Caused By..."] + bt }
+        bt_out = bt_lines.join "\n    "
+        "#{self.exception.class}: #{self.exception.message}\n    #{bt_out}"
+      else
+        bt = Minitest.filter_backtrace(self.backtrace).join "\n    "
+        "#{self.exception.class}: #{self.exception.message}\n    #{bt}"
+      end
     end
 
     def result_label # :nodoc:
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -263,6 +263,40 @@ class TestMinitestRunner < MetaMetaMetaTestCase
     assert_report expected
   end
 
+  def test_run_nested_error
+    @tu =
+    Class.new FakeNamedTest do
+      def test_something
+        assert true
+      end
+
+      def test_error
+        begin
+          raise "internal exception"
+        rescue
+          raise "external exception"
+        end
+      end
+    end
+
+    expected = clean <<-EOM
+      E.
+
+      Finished in 0.00
+
+        1) Error:
+      FakeNamedTestXX#test_error:
+      RuntimeError: external exception
+          FILE:LINE:in \`rescue in test_error\'
+          FILE:LINE:in \`test_error\'
+          FILE:LINE:in \`test_error\'
+
+      2 runs, 1 assertions, 0 failures, 1 errors, 0 skips
+    EOM
+
+    assert_report expected
+  end
+
   def test_run_error_teardown
     @tu =
     Class.new FakeNamedTest do


zenspider:
I'm actually a bit torn about this as it might hold onto an exception that we can't marshal. Cc: @tenderlove

noahgibbs:
If I do a quick test using Marshal, it looks like inner exceptions are Marshal-able. So I think you're good on that.

Specifically, in my local Ruby 2.3.1:

```
begin
  begin
    raise "Inner message"
  rescue
    raise "Outer message"
  end
rescue
    nest_e1 = $!
end

 => #<RuntimeError: Outer message>
2.3.1 :007 > nest_e1
 => #<RuntimeError: Outer message>
2.3.1 :008 > nest_e1.cause

nest_me1 = Marshal.dump nest_e1
nest_ne1 = Marshal.load(nest_me1)

2.3.1 :011 > nest_ne1.cause
 => #<RuntimeError: Inner message>
```

zenspider:
Not if the inner exception isn’t marshalable. Tho it looks like right now I’m not insulating myself from that anyways… It is something we need a test for so we can ensure it is addressed. Probably needs to go into a bug release shortly.

cc/ @tenderlove

noahgibbs:
That's true. Your old approach (pre-nested-exceptions) had you replacing a non-Marshalable exception with a Marshalable one. You could replace the "error" object with one not created inside a rescue block to avoid the old one being included as "cause". My understanding is that you can't explicitly set or clear cause for an exception except by creating it in a rescue block or not - it automatically picks up dollar-bang as the value of "cause".

zenspider:
Mmmm... `$! = nil` before creating the new exception?

This is going to be a bitch to test...

noahgibbs:
There is a keyword to "raise" that will let you specify the inner exception, including as nil. That may be your best bet.

The raise keyword doesn't seem to be documented where I can find it, so I wrote it up: "http://engineering.appfolio.com/appfolio-engineering/2017/12/28/ruby-and-nested-exceptions" (scroll down to "Secrets.")

lib/minitest.rb
```diff
@@ -836,8 +836,24 @@ def error # :nodoc:
     end
 
     def message # :nodoc:
-      bt = Minitest.filter_backtrace(self.backtrace).join "\n    "
-      "#{self.exception.class}: #{self.exception.message}\n    #{bt}"
+      if self.exception.respond_to?(:cause) && self.exception.cause
+        # Build a chain of exception causes starting from the outermost
+        exc = self.exception
```
fsateler, line 841: This has been renamed to `error` with https://github.com/seattlerb/minitest/commit/5802aa08280a3609b684e0ebef95dabfbd3ec9ff

fsateler:
This has the slightly annoying problem of duplicating a lot of trace lines, because the inner exception trace contains a large chunk of the outer exception trace.

Nakilon:
I was asked to come up with tests for my patch I did here https://github.com/Nakilon/minitest/commit/b7498b0b08f7afe6e64fdd71f1e8f60a874bd148 about the same issue but I'm not ready for that. You guys may consider looking at it.
Yes, I saw the backtrace duplication too -- maybe it should just prefer the cause's backtrace over another one, idk.

ianwhite:
I've been using the following patch in my tests to show error causes.  I've moved the marshallable/sanitize logic into the UnexpectedError class, and only save the error details for the error and its causes.  This ensures UnexpectedError is marshallable, and also has the effect of showing the original error (instead of the wrapped error) in the error message.  Only backtrace lines not already shown are shown in the cause backtraces.

Let me know if this is still of interest, and I can work on a pull request

```ruby
# enhance Minitest to show exception causes
module MiniTest
  class Test
    def capture_exceptions # :nodoc:
      yield
    rescue *PASSTHROUGH_EXCEPTIONS
      raise
    rescue Assertion => e
      self.failures << e
    rescue Exception => e
      self.failures << UnexpectedError.new(e)
    end
  end

  class UnexpectedError
    attr_accessor :error, :details

    def initialize error # :nodoc:
      super "Unexpected exception"
      self.error = sanitize_exception(error)
      self.details = error_details(error)
    end

    def message
      seen_bt = []
      details.map do |d|
        bt = Minitest.filter_backtrace(d.backtrace)
        bt -= seen_bt
        seen_bt |= bt
        "#{d.error_class}: #{d.message}\n    #{bt.join "\n    " }"
      end.join("\n  - Cause: ")
    end

    private

    def sanitize_exception(e)
      Marshal.dump e
      e
    rescue TypeError
      bt = e.backtrace
      e = RuntimeError.new "Wrapped undumpable exception for: #{e.class}: #{e.message}"
      e.set_backtrace bt
      e
    end

    def error_details(error)
      ([error] + causes(error)).map { |e| ErrorDetail.new(e) }
    end

    def causes(e)
      (e.respond_to?(:cause) && e.cause) ? [e.cause] + causes(e.cause) : []
    end

    class ErrorDetail # :nodoc:
      attr_reader :error_class, :message, :backtrace

      def initialize(error)
        @error_class = error.class
        @message = error.message
        @backtrace = error.backtrace
      end
    end
  end
end
```

It produces output like this:

```
Error:
UserAuthenticationTest#test_login_with_inivaid_credentials,_request_reset_password,_reset_password,_log_in,_log_out:
ArgumentError: wrong number of arguments (given 1, expected 0)
    app/repos/user_repo.rb:9:in `block in <class:UserRepo>'
    app/repos/user_repo.rb:12:in `authenticate'
    app/actions/users/logins/create_action.rb:8:in `set_result'
    app/controllers/users/logins_controller.rb:8:in `create'
    app/utilities/result_error_filter.rb:4:in `around'
  - Cause: UserRepo::NotAuthenticatedError: UserRepo::NotAuthenticatedError
    app/repos/user_repo.rb:14:in `block (2 levels) in authenticate'
    app/repos/user_repo.rb:13:in `block in authenticate'
```

zenspider:
I've spent too much time wrangling unmarshallable (that's a word, right?) exceptions at ONE level to be comfortable with this. 

This idea needs comprehensive tests specifically around a chain of `cause`s where one or more up the chain will cause a TypeError. It's hairy AF.

Unfortunately, while `raise` has the `cause` keyword, `Exception.new` does not and `$!` is read-only, which means I can't readily construct new exceptions that don't have `cause` hooked up... so this might already be biting us. Until we have tests that cover this edge case, we'll never really know.

I'm going to close this. I am open to reinvestigating this once the tests cover a more comprehensive matrix of edge cases.

PR #728, opened 2017-12-25
[ci skip] Document how to run tests via Rake

schneems:


utilum:
Minitest's own tests suggest  `t.test_files = FileList["test/**/test_*.rb"]`

utilum:
Perhaps noting that options then need to be passed in `TESTOPTS` ?

zenspider:
I'm gonna add this... and it's gonna last about 2 weeks. After that, I'm adding `minitest/test_task.rb` and replacing the doco for that. Rake::TestTask is an unusable abomination.

zenspider:
Done. Thank you. Sorry I took so long to tend to this.

PR #733, opened 2018-01-03
Fix class name in deprecation message

y-yagi:
Because this method defined at `Runnable` class. Also, `marshal_dump` is
undefined at `Result` class. So, there is nothing called from `Test`.


zenspider:
Done. Thanks!

PR #735, opened 2018-01-10
Location does not always returns the class name

alexisbernard:
When I call `location` from a test, it simply returns `#method` instead of `Class#method`. However it used to work properly in 5.10.3.

```ruby
require "minitest/autorun"

class FooTest < Minitest::Test
  def test_bar
    assert_equal("FooTest#test_bar", location)
  end
end
```

```
$ ruby test_foo.rb
  1) Failure:
FooTest#test_bar [foo_test.rb:5]:
--- expected
+++ actual
@@ -1,2 +1,2 @@
 # encoding: UTF-8
-"FooTest#test_bar"
+"#test_bar"
```

### DIFF
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -85,6 +85,7 @@ module Minitest
     # Runs a single test with setup/teardown hooks.
 
     def run
+      self.klass = self.class
       with_info_handler do
         time_it do
           capture_exceptions do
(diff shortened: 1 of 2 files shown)


zenspider:
Curious… but why are you trying to use `location` in a test?

zenspider:
This is fixed. Thanks for reporting (I’d still like that question above answered if you can)

alexisbernard:
Indeed, that is not common to call `location` from a test method :-) It's convenient when using VCR :

```ruby
def test_foo
  VCR.use_cassette(location) do
    ...
  end
end
```

Issue #616, opened 2016-04-07
Quiet Minitest::Benchmark output


ghost:
I've incorporated Minitest::Benchmark into my the primary rake :test task that runs all tests (coverage, feature, unit, and performance). This get's run primarily for CI purposes, but occasionally manually in development. When I do run it for development purposes I like to keep the output as clean as possible (just a bunch of ... success dots FF failures etc.) and Minitest::Benchmark was giving output every time.

I dove into the code and saw I could overwrite the reporter with IO to /dev/null. Here is that [nasty hack](https://github.com/wurde/canvas/blob/a602e7a318ea11bfc9c37d65f1f2d995ecb9ba6c/lib/canvas/ruby/canvas_test_performance.rb#L284-L288) to redirect output to /dev/null.

PuZZleDucK:
I've failed in a naive attempt to add this to MiniTest... however there was a suggestion that this might be a good candidate for a reporter plugin.

PuZZleDucK:
I've not been able to work out how to even run a plugin... far from writing one. I'm afraid this one is over my head.

utilum:
minitest-reporters ?

PR #745, opened 2018-02-11
add --quiet flag

PuZZleDucK:
This patch adds a command line flag ("-q" or "--quiet") to suppress all output other than the single character pass/fail/skip result.

This address issue #616

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -287,6 +287,7 @@ provided via plugins. To see them, simply run with +--help+:
         -h, --help                       Display this help.
         -s, --seed SEED                  Sets random seed. Also via env. Eg: SEED=n rake
         -v, --verbose                    Verbose. Show progress processing files.
+        -q, --quiet                      Quiet. Show minimal output.
         -n, --name PATTERN               Filter run on /regexp/ or string.
         -e, --exclude PATTERN            Exclude /regexp/ or string from run.
 
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -186,6 +186,10 @@ module Minitest
         options[:verbose] = true
       end
 
+      opts.on "-q", "--quiet", "Quiet. Show minimal output." do
+        options[:quiet] = true
+      end
+
       opts.on "-n", "--name PATTERN", "Filter run on /regexp/ or string." do |a|
         options[:filter] = a
       end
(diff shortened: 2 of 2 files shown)


utilum:
"Quiet" is open to interpretation, and this is one. Personally I'd expect it to mean only the aggregate results, the dots can span more than a screenful.

PuZZleDucK:
Good point @utilum ... could also mean no output whatsoever (which was the old definition of "-q"). I know I've redirected to null and just examined the return value in CI before, and I've read of others doing similar.

We could do some kind of -qqq thing where:
-q prints dots and summary
-qq prints summary
-qqq no output

I think this might be a bigger change and warrant refactoring the print lines to remove the verbosity checks everywhere... I already thought this patch was pushing it.

zenspider:
I don’t like this and don’t see the point. Here is what minitest looks like normally (with a fake failure injected):

```
Run options: --seed 64160

# Running:

......................................................................................................................................................................................................................................................................................................F......S....S.SS..S..S.S.S..S........S......

Finished in 0.165647s, 2040.4837 runs/s, 5698.8657 assertions/s.

  1) Failure:
TestMinitestUnit#test_no [/Users/ryan/Work/p4/zss/src/minitest/dev/test/minitest/test_minitest_test.rb:62]:
Epic Fail!

338 runs, 944 assertions, 1 failures, 0 errors, 10 skips

```

and with this option on:

```
....................................................................................................................................................................................................................................................................F........................................SS...SS..SS..S.S..S....S.............
  1) Failure:
TestMinitestUnit#test_no [/Users/ryan/Work/p4/zss/src/minitest/dev/test/minitest/test_minitest_test.rb:62]:
Epic Fail!
```

Now you’re missing information that would allow you to reproduce these results, the formatting seems less readable, and I’m not sure what the benefit is or why someone would want it. Can someone clarify this part?

Finally, this doesn’t address #616... it seems to be the opposite of it:

```
9998 % ruby -Ilib bench.rb -q
bench_my_algorithm3	 0.000006	 0.000002	 0.000002	 0.000002	 0.000001
.bench_my_algorithm4	 0.000003	 0.000002	 0.000002	 0.000001	 0.000002
.bench_my_algorithm5	 0.000003	 0.000002	 0.000001	 0.000002	 0.000002
.bench_my_algorithm1	 0.000003	 0.000001	 0.000001	 0.000002	 0.000002
.bench_my_algorithm2	 0.000004	 0.000002	 0.000002	 0.000002	 0.000002
.
```

utilum:
@PuZZleDucK : Have you thought of writing a plugin for this?

PuZZleDucK:
@zenspider silly me... I only looked at the happy-path and didn't consider what got printed on a failure :(

My personal use case for something similar was for a CI setup where the tests had already been run earlier in the process. The final run was just a sanity check before deployment and it was nice to have each test suite only print one line so it was all easy to see on one page.

@utilum no... no I haven't... but I should!
A quiet reporter plugin sounds like a great fit. 👍

PuZZleDucK:
I'm afraid the plugin thing was a bit over my head... after a few days of trying I've not even been able to include or run a plugin... or understand how I'm meant to be including or running it :p

If there is a super simple plugin example someone could suggest I'll try again and give it another go, but I'm so lost on how to do it right now.

PR #751, opened 2018-03-28
Add minitest.gemspec

### LINKED ISSUE #750: Install from git
If this repo had `minitest.gemspec` file in it, then we could install latest version like this:
```ruby
gem 'minitest', github: 'seattlerb/minitest'
```

But it doesn't have it :(

Paxa:
Fixes #750

I removed `Manifest.txt` from git because it's not necessary anymore, but it will be generated whenever you call rake (still not sure why hue needs it)

Do you think we can remove all rake tasks related to creation of gem? (usually `gem build minitest.gemspec && gem push minitest-5.11.3.gem` would be enough)

Also shortened description, because current one is too long
<details>
  <summary>description screenshot</summary>
  screen shot 2018-03-28 at 16 04 07
</details>


minitest.gemspec
```diff
@@ -0,0 +1,50 @@
...
+                     "minitest/test and seamlessly bridges test assertions over to spec expectations.\n"\
+                     "minitest/benchmark is an awesome way to assert the performance of your\nalgorithms "\
+                     "in a repeatable manner. Now you can assert that your new co-worker doesn't replace your "\
+                     "linear algorithm with an exponential one!\n" \
+                     "nminitest/mock by Steven Baker, is a beautifully tiny mock (and stub) object framework.\n"\
+                     "minitest/pride shows pride in testing and adds coloring to your test output. "\
+                     "I guess it is an example of how to write IO pipes too."
+
+  spec.require_paths    = ["lib"]
+  spec.files            = `git ls-files`.split("\n")
```
SuperSandro2000: Why not put the test files into ``spec.test_files``?
``spec.files = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(lib)/}) }``

minitest.gemspec
```diff
@@ -0,0 +1,50 @@
...
+  spec.test_files       = []
```
SuperSandro2000: ``spec.test_files = `git ls-files -z`.split("\x0").select { |f| f.match(%r{^(test)/}) }``

Paxa: Ok, why do you use `-z` for ls-files ?

SuperSandro2000: ``git ls-files -z lib
lib/canuby.rb lib/util.rb
``

``git ls-files lib
lib/canuby.rb
lib/util.rb

Paxa: To handle situation when file name contain `\n`?

SuperSandro2000: With the ``-z`` flag the files are printed in one line. Without it each file gets a new line. Removing the ``-z`` flag would require to change the character in split.

Paxa: As I understand, the result will be same right?
```
`git ls-files -z`.split("\x0") == `git ls-files`.split("\n")
```

SuperSandro2000: yes

zenspider:
> I removed Manifest.txt from git because it's not necessary anymore, but it will be generated whenever you call rake (still not sure why hue needs it)

That is not how I work. Manifests are important to a project, as evidenced by how much you two have discussed various ways in which `git ls-files` can mess things up. `git add -a` is a perfect example as to why manifests are valuable.

See also: http://www.zenspider.com/~ryan/Hoe.pdf
and: http://www.zenspider.com/ruby/2008/08/do-the-simplest-thing-that-could-possibly-work-applies-to-more-than-code.html

> Do you think we can remove all rake tasks related to creation of gem? (usually gem build minitest.gemspec && gem push minitest-5.11.3.gem would be enough)

No, I do not think that. That's not how I work.

Please don't wade into a project you have no prior contributions to and completely upend the way the development process is done. At least discuss it with the owners and/or maintainers first.

Paxa:
> Please don't wade into a project you have no prior contributions

Actually I do

>  `git add -a` is a perfect example as to why manifests are valuable.

Can you provide some example? Aren't people usually do `git status` before commit? (I don't know about others because I use for GUI to review changes and write commit message)

What I see, most of projects moved away from generating `*.gemspec` files by other tools and maintain it manually. I do the same for my projects and I find it convenient (transparent, support install from git, well known in community). As you mentioned several times, that's not how you work, got it!

My proposal is about making minitest runnable from git, if you agree with this idea - I will implement it according to how you work 

---

> http://www.zenspider.com/ruby/2008/08/do-the-simplest-thing-that-could-possibly-work-applies-to-more-than-code.html

I guess it's about personal preferences, in general I like reducing unnecessary code and files. But adding `Manifest.txt` (which is always same with `git ls-files`), reduce usability and adding dependency that does a lot of staff - seems a bigger evil for my needs (build gem, publish gem, generate docs). I guess we are looking from different perspectives because hoe is 90% done by you... btw I noticed that projects are usually using `CHANGELOG.md` now days

SuperSandro2000:
I don't really see the point of hoe either but if you like it 🤷‍♂️. I could imagine using it with very very very big projects (kubernetes size) but in smaller projects it abstracts the build process and makes it harder for new people to understand what is going on.
I don't currently do no plan to do minitest contributions so I don't really care and only looked at his changes to be right not ethical correct.

zenspider:
> Can you provide some example?

Easy: https://github.com/search?q=.DS_Store&type=Commits&utf8=✓

750k commits, 19k issues... and so on.

> My proposal is about making minitest runnable from git, if you agree with this idea - I will implement it according to how you work.

Again, I don't. I release minitest early and often. 114 releases so far. I don't support running off of master because I don't guarantee its usability until I release it.

> I don't really see the point of hoe

You probably don't have 90+ gems and contribute to many many more.

> in general I like reducing unnecessary code and files

that is exactly the point of Hoe.

Paxa:
How do you usually maintain `Manifest.txt`?
do `git status` before `git commit` or see added files when writing a commit message in text editor?
Have you event published a gem with missing file because `Manifest.txt` wasn't updated?

utilum:
@Paxa try looking at the output from `rake -T`, and maybe at Hoe tasks more closely.

Paxa:
Does it mean 90+ gems can not be installed from git?

@utilum I did, and I found `rake check_manifest`, does Hoe's users suppose to run it before git push?

btw I notice that most of projects in https://github.com/seattlerb don't have `.gitignore` file, is there any reason behind it?

utilum:
I suspect it may be related to the developers' choice of a revision control system, perhaps that even makes them projects more portable, being version control system agnostic.

utilum:
@Paxa see https://www.zenspider.com/~ryan/Hoe.pdf.
And now I feel like I've played the role of search engine for you enough for a period.

SuperSandro2000:
I can understand that if you have a lot of gems updating common things can be a pain. Creating the gemspec in hoe is a clever idea but the lack of documentation makes it a pain for new users to understand that. Not including a gitignore on the other hand is just :shrug:. And doing things version control independent now days is like: why don't you ever want to use git? The project is old and if you would do a new one now days you would do a lot of things different. Maybe don't create 88 gems in the first place. This discussion is pointless and changing old projects significantly without a good reason is never a good idea. 
PS: that graph in the hoe.pdf is not representing anything because a lot more factors changed in the time except the addition of hoe.
And FYI hoe is slang for prostitute in Germany.

PR #765, opened 2018-06-10
Print '--seed=1234' rather than '--seed 1234' by default

cben:
before:
    Run options: --seed 62153
after:
    Run options: --seed=10759

This is friendlier to passing options via Rake.  These forms work:
    env TESTOPTS='--seed=1234' rake test
    rake test TESTOPTS='--seed=1234'
These forms with a space fail for me:
    env TESTOPTS='--seed 1234' rake test
    rake test TESTOPTS='--seed 1234'
with message like:
    File does not exist: /home/me/kubeclient/1234

Of course this assumes one even discovers the rake `TESTOPTS` var...
Still, one less obstacle :-)

(I see from #564 you're not interested in minitest directly telling user Rake-specific hints how to use `TESTOPTS`.
Consider perhaps printing `SEED=1234` hint instead of `--seed` hint?  This works when running minitest directly, through Rake and probably through other runners.)

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -220,7 +220,7 @@ module Minitest
     unless options[:seed] then
       srand
       options[:seed] = (ENV["SEED"] || srand).to_i % 0xFFFF
-      orig_args << "--seed" << options[:seed].to_s
+      orig_args << "--seed=#{options[:seed]}"
     end
 
     srand options[:seed]


zenspider:
I gotta say... I absolutely _hate_ Rake's TestTask. I think it is just about the least usable thing out there. How about I replace it entirely with something that works well rather than applying bandaids to deal with rake?

https://github.com/seattlerb/hoe/blob/master/lib/minitest/test_task.rb

(this will be moved to minitest after running some laps)

cben:
Friendly ping :wave:
Any progress with replacing TestTask?  IIUC the code you linked that would provide `Minitest::TestTask`, which will be opt-in, won't affect all the projects that followed current README's advice to use `Rake::TestTask`.
Any reason not to merge this tiny change?

PR #766, opened 2018-06-13
Correctly unescape newlines in diffs

tomstuart:
@carolineschnapp noticed that, if the code under test mistakenly uses a single-quoted string rather than a double-quoted one, the diff shown by `assert_equal` is unhelpful in the presence of newlines:

```
mistake = 'hello\nworld'
assert_equal "hello\nworld", mistake

Failure:
Test#test_assert_equal:
--- expected
+++ actual
@@ -1,2 +1,2 @@
-"hello
+"hello\
 world"
```

Whereas the expected diff would be:

```
Failure:
Test#test_assert_equal:
--- expected
+++ actual
@@ -1,2 +1 @@
-"hello
-world"
+"hello\\nworld"
```

This is happening because of indiscriminate unescaping of the character sequence “\n” in `mu_pp_for_diff`. In the case where the backslash is itself escaped, as in the string `"hello\\nworld"`, it represents a literal backslash and therefore should have no effect on the subsequent “n”.

In lieu of `String#undump` [1,2] being universally available to solve the more general problem, our best bet is to use a more permissive pattern in `String#gsub` to only pick out escape sequences which aren’t already part of another escape sequence, then replace those sequences if they happen to represent a newline.

[1] https://github.com/ruby/ruby/commit/bbec11d329f5a72fe6151ec9fb0e25ff255f2eed
[2] https://github.com/tadd/string_undump

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -125,7 +125,7 @@ module Minitest
     # uses mu_pp to do the first pass and then cleans it up.
 
     def mu_pp_for_diff obj
-      mu_pp(obj).gsub(/\\n/, "\n").gsub(/:0x[a-fA-F0-9]{4,}/m, ":0xXXXXXX")
+      mu_pp(obj).gsub(/\\./) { |m| m == '\n' ? "\n" : m }.gsub(/:0x[a-fA-F0-9]{4,}/m, ":0xXXXXXX")
     end
 
     ##
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -1055,6 +1055,21 @@ class TestMinitestUnitTestCase < Minitest::Test
     end
   end
 
+  def test_assert_equal_unescape_newlines
+    msg = <<-EOM.gsub(/^ {10}/, "")
+          --- expected
+          +++ actual
+          @@ -1,2 +1 @@
+          -"hello
+          -world"
+          +"hello\\\\nworld"
+          EOM
+
+    assert_triggered msg do
+      @tc.assert_equal "hello\nworld", 'hello\nworld'
+    end
+  end
+
   def test_assert_equal_does_not_allow_lhs_nil
     if Minitest::VERSION =~ /^6/ then
       warn "Time to strip the MT5 test"


zenspider:
I'm not sure I agree with your proposal and I'm wrestling with trying to distill it down. I prefer the current behavior to yours as it makes it easier for a failing diff to show you where the problem is. The trailing `\` is intentional, to show you that the newline is escaped. My current counter-proposal is to make a trailing `\n` _and_ an actual newline. I think this balances out by increasing visibility with the current improved diffs.

What I have done so far is to switch to testing `mu_pp` and `mu_pp_for_diff` directly to show the current intent, your proposal, and my current counter-proposal. For me, it is easier to read the diffs:

https://gist.github.com/zenspider/4914e14f637e691ec77a518274370801/revisions

but you might prefer the final product:

https://gist.github.com/zenspider/4914e14f637e691ec77a518274370801

Thoughts?

zenspider:
I just bumped the gist to include your higher level test w/ modified for my counter-proposal.

zenspider:
I think your summary is accurate, but your focus on "round-trippable" strings (that's a word, right?) is confusing to me. There's nothing about `mu_pp_for_diff` that purports to be round-trippable. In fact, it is there **to mutate** *in order to make diffs more readable*. 

Breaking up lines (admittedly arbitrarily) by literal newline is just one way to improve readability of a (long) diff. As another counterproposal, I could totally see this as being a (somewhat) valid implementation:

```ruby
def mu_pp_for_diff obj
  mu_pp(obj).gsub(/ /, " \n") # or /\b/
end
```

such that seeing a diff from this output will tell you via line-number exactly which word the diff starts at. That would make it trivial for me to jump to the difference in my editor (emacs) with `M-### M-f` (jump forward by word `###` times). But it would be a pain to read and you wouldn't have enough context surrounding the diff to be useful.

Where I agree that there was a bug in the current implementation is that I was destroying/obscuring information that made it harder to distinguish between your `string1` and `string2` examples. I think my counterproposal balances [damnit github... you destroyed half my writeup...] um... balances readability versus comprehension.

I do agree that this might boil down to a difference in opinion/style here. I'm resolute that "readable diffs" are an important feature for `assert_equal`, but I'm open to the ways in which they can be improved.

Does this make sense? Half my edits were dropped by github and I'm not sure I restored the whole thing.

zenspider:
Ping. I'm going to go forward with my counter-proposal without further feedback. I'd love to hear back from you to see if I've made sense on this one.

tomstuart:
I think you’re right that we just have differing opinions on style here. I feel that the newline insertion in your counter-proposal has the potential to be confusing for someone who’s trying to debug a newline-related failure, but that’s not a big deal in the context of the issue I originally reported. Your solution does address the underlying bug, so I’m happy!

zenspider:
Kk... Thank you. I'll see what I can do to improve the doco around this as well. I'll get the commit in shortly. Leaving open until I do because I'm pretty off today.

zenspider:
I _finally_ got this in last night. A couple iterations actually. I think the end result is the right balance.

Sorry this took me so long to get back on this.

tomstuart:
Not at all — thank you very much for addressing it!

PR #770, opened 2018-07-09
Taking out global expectations from README

CheezItMan:
I just think if global expectations are being depreciated, it should be removed from the project README.

Very minor suggested change

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -126,13 +126,13 @@ Define your tests as methods beginning with +test_+.
 
     describe "when asked about cheeseburgers" do
       it "must respond positively" do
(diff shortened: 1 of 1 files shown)


cpb:
You may be more successful with this pull request if you changed it from `expect` to `_` or `value` as zenspider indicated a preference for the later two options in the railsconf talk about minitest 6.

[I think it is terrible but \[expect\] is there too](https://youtu.be/l-ZNxvFo4lw?t=2084)

zenspider:
Sorry for taking so long. I'm gonna apply but switch to `_` instead of `expect`.

zenspider:
done. thanks!

PR #772, opened 2018-08-12
Improve documentation for stubbing with callables

### LINKED ISSUE #771: Insufficient determination between val_or_callable
This issue appeared in Version: 5.10.3

As described in [this Stack Overflow question](https://stackoverflow.com/questions/51311387/minitest-stub-passing-block-to-mock-instance), I have encountered a problem while trying to stub a method call to provide an instance of [MockRedis](https://github.com/brigade/mock_redis). Minitest passed the provided `do-block`, in which the stub should be used, as argument to the MockRedis instance.

The problem lies in the following [code block](https://github.com/seattlerb/minitest/blob/e6bc4485730403faff6966c1671cf5de72b2d233/lib/minitest/mock.rb#L226), because `MockRedis` responds to `:call`:

```ruby
ret = if val_or_callable.respond_to? :call then
        val_or_callable.call(*args)
      else
        val_or_callable
      end
```

alex-kuck:
This is a proposal to improve the documentation for the default behaviour when using stubs that respond to `call`.

This closes #771.


cpb:
I think you might have more success with this pull request if you remove the .gitignore file.

Your proposed documentation change is a bit confusing. The way you phrase "the stubbed instance" makes it unclear if it is the object receiving the call to `stub` or the substitute return value. In the documentation zenspider references, it uses the argument name. It shows that `+val_or_callable+` is checked for `call` implementation.

Blocks and stubs or mocks are a tricky situation. How does your use case work with the minitest 6 branch? Have you tried `stub6` or [`stub6_2`](https://github.com/zenspider/minitest/commit/91db4d8ebf95e941f2d88bcc6b64ccdd4e7837e9#diff-00a18d559ad9a1e984068e38e5009047R259)?

The solution might not be in documentation alone!

zenspider:
Checking back in... I agree. The language (as well as the idea being described) is awkward and confusing. For example, `Minitest` isn't checking for a callable `stub` is. I think it would be clearer to actually use `lambda` since that's what you call it. Etc.

and yes, no .gitignore, and 80 columns please, (tho I can reformat to my liking when I pull this in manually)

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
zenspider: no

PR #778, opened 2018-09-26
Allow nil on the left-hand side of assert_equal IF explicit

laserlemon:
The caller must pass the `allow_nil: true` option or else (starting with Minitest 6), the assertion will fail if a LHS `nil` value is encountered.

## :warning: Warning… ☠️🐴 alert!

 I know that this topic has been [discussed at length](https://github.com/seattlerb/minitest/issues/666), but I'm new to the conversation and noticed that one idea that hadn't come up (at least that I'd seen) was to allow for an _explicit_ opt-in to the otherwise dangerous LHS `nil` behavior.

This would allow a developer to extend Minitest with equality assertion helpers, but without the `nil` value check conditionals. In cases where the developer certainly wants this behavior, it's available. In typical usage, safe practices are encouraged/enforced.

I don't have a :dog: in the fight so I'm still happy whether this is merged or closed. Mostly, I'd never contributed to Minitest and wanted to. :smile: I tried to mimic the code style and hope I did it justice!

I do hope that this change might be a reasonable middle ground solution! Thank you for your ongoing and great work on Minitest! 🏆💐

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -169,11 +169,18 @@ module Minitest
     #
     # See also: Minitest::Assertions.diff
 
-    def assert_equal exp, act, msg = nil
+    def assert_equal exp, act, msg_or_opts = nil, opts = {}
+      if Hash === msg_or_opts then
+        msg = nil
+        opts = msg_or_opts
+      else
+        msg = msg_or_opts
+      end
+
       msg = message(msg, E) { diff exp, act }
       result = assert exp == act, msg
 
-      if nil == exp then
+      if nil == exp && !opts[:allow_nil] then
         if Minitest::VERSION =~ /^6/ then
           refute_nil exp, "Use assert_nil if expecting nil."
         else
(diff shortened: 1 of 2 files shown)


cpb:
Tenderlove makes the same point last year: https://github.com/seattlerb/minitest/issues/666#issuecomment-311500552

zenspider:
I appreciate the PR and the polite tone. Thank you for trying to contribute. But this isn't going to happen. The people arguing against this change to `assert_equal` are being beyond hyperbolic when they say they're going to be entirely put out by this. If they're generating tests or writing their own assertions, they've already written 5+ lines of code. Adding an `if` won't hurt anything.

I'm going to leave this open for a short time in case you want to respond, but I suspect I will lock this down as soon as the hyperbolic detractors show up. I'm quite tired of the attitude and it has put me off OSS for a while.

laserlemon:
@cpb @zenspider Thank you for your responses! I'm happy to close this myself. I certainly don't want to revive a painful (and closed) conversation. Again, thank you for your amazing work all around!

laserlemon:
💛

zenspider:
Thank you. Locking preemptively.

PR #788, opened 2019-04-02
Kind error message for missing block

okuramasafumi:
`assert_output` and `assert_raises` requires a block to work.
Currently, without block, it fails with `LocalJumpError`,which is not helpful.
Those assertions now fail with kind error message without a block.

If you found the message awkward, please modify it.


zenspider:
I just moved all the assertion tests so this won't apply cleanly. I'll hand merge.

zenspider:
Done, with some minor rewording. Thank you!

PR #790, opened 2019-04-25
Add path assertions

deivid-rodriguez:
This PR is a follow up to https://github.com/rubygems/rubygems/commit/f33345d6e62af339299d873d03cd3cfd14f9c989#diff-0c4b3e9996ebec38f191e68264665984R82.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -442,6 +442,14 @@ module Minitest
     end
 
     ##
+    # Fails unless +path+ exists.
+
+    def assert_path_exists(path, msg = nil)
+      msg = message(msg) { "Expected path '#{path}' to exist" }
+      assert File.exist?(path), msg
+    end
+
+    ##
     # Captures $stdout and $stderr into strings:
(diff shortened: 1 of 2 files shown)


bronzdoc:
👀

lib/minitest/assertions.rb
```diff
@@ -564,7 +572,7 @@ def refute test, msg = nil
     end
 
     ##
-    # Fails if +obj+ is empty.
+    # Fails if +obj+ is not empty.
```
zenspider: Not quite. This is either "Fails if obj is empty" or "Fails unless obj is not empty".

Think of it backwards: it's the opposite of `assert_empty` which "fails if obj is non-empty".

zenspider:
Added and rearranged. Thank you!

lib/minitest/assertions.rb
deivid-rodriguez: Right, I don't know what I was thinking :/. Thanks for getting this in.

PR #792, opened 2019-05-30
Refactor self.need_to_diff?

BurdetteLamar:
The code for whether need to diff can (and does) behave differently on different platforms.

In my testing, I need to know whether a failed assertion will diff, because the 'actual' value will vary accordingly.  Therefore I'm proposing to expose the logic as a new method, self.need_to_diff?

Minitest's own tests, before and after this change, pass on my Linux box.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -44,6 +44,19 @@ module Minitest
     end
 
     ##
+    # Returns whether need to diff.
+    def self.need_to_diff? expect, butwas
+          (expect.include?("\n")    ||
+              butwas.include?("\n")    ||
+              expect.size > 30         ||
+              butwas.size > 30         ||
+              expect == butwas)        &&
+              Minitest::Assertions.diff
+
+
+    end
+
+    ##
     # Set the diff command to use in #diff.
 
     def self.diff= o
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
Not the best solution.

zenspider:
Besides the misformatting... what was wrong with your PR? Why did you close?

BurdetteLamar:
To call the new method, I would need both expect and butwas.  These are computed in self.diff, and cannot be moved b/c they are still needed there.  So more change would be required to make this right, I think.

Not worth the additional trouble?

Thanks for considering, though.

BurdetteLamar:
Oh, and sorry for misformatting.  The hundreds of PRs I've created before have all been for my own projects, and so I was not mindful of being a good PR citizen.  Point taken.

zenspider:
Hrm... Would it be too weird for the refactoring to calculate and return the things that need diffing rather than a plain boolean?

BurdetteLamar:
You're thinking that the new method would accept  (exp, act) and return its computed [expect, butwas] or nil?

zenspider:
Yup. Prolly needs a different name, like `things_to_diff`. Prolly should contain your current refactoring too?

BurdetteLamar:
PR 793 is up.  Thx for  encouragement!

PR #793, opened 2019-06-06
Refactor to form method things_to_diff

BurdetteLamar:


BurdetteLamar:
Allow the test to discover whether diffing will be done.
This matters because the decision is platform-dependent, and therefore the assertion's message will also be platform-dependent.

zenspider:
> This matters because the decision is platform-dependent, and therefore the assertion's message will also be platform-dependent.

Can you explain this part a bit more? Just curious...

BurdetteLamar:
Sure.  I'm working on a project whose tests use minitest.  I'd like for the app (it's a Ruby gem) to work on both Windows and Linux.  I sometimes get a different result, depending on the platform, so I'd like to know whether the diffing will be done.

(I know that the choice of differ is also platform-dependent, but in my case I already have Gem diff-lcs available, so I've coerced minitest to always use that..)

zenspider:
> depending on the platform, so I'd like to know whether the diffing will be done.

oh... because you don't have `diff.exe` on windows or simply because you'd rather use `diff-lcs`? Makes sense. I suspect your tests might get a bit painful if you have a change that causes a lot of failures. It's been a while so it might have gotten better, but diff-lcs very clearly lost out to unix `diff` even with the extra shell out and capture.

BurdetteLamar:
> because you don't have diff.exe on windows or simply because you'd rather use diff-lcs?

I'm already using diff-lcs in my testing (for other reasons), so I can force its use in minitest, instead of letting minitest make the choice.

BurdetteLamar:
Anything more I need to do with this PR?

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
...
+      expect = mu_pp_for_diff exp
+      butwas = mu_pp_for_diff act
+      need_to_diff =
+          (expect.include?("\n")    ||
+              butwas.include?("\n")    ||
+              expect.size > 30         ||
+              butwas.size > 30         ||
+              expect == butwas)        &&
+              Minitest::Assertions.diff
+      need_to_diff ? [expect, butwas] : [nil, nil]
```
zenspider: this might just be me... but I would do `need_to_diff && [expect, butwas]` and let ruby do the rest

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
...
+    def things_to_diff(exp, act)
+      expect = mu_pp_for_diff exp
+      butwas = mu_pp_for_diff act
+      need_to_diff =
+          (expect.include?("\n")    ||
+              butwas.include?("\n")    ||
+              expect.size > 30         ||
+              butwas.size > 30         ||
+              expect == butwas)        &&
+              Minitest::Assertions.diff
```
zenspider: ditto on blank line

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
       @diff = o
     end
 
+    ##
+    # Returns things to diff [expect, butwas], or [nil, nil] if nothing to diff.
+    def things_to_diff(exp, act)
+      expect = mu_pp_for_diff exp
+      butwas = mu_pp_for_diff act
+      need_to_diff =
+          (expect.include?("\n")    ||
```
zenspider: indentation is funny here.

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
       @diff = o
     end
 
+    ##
+    # Returns things to diff [expect, butwas], or [nil, nil] if nothing to diff.
+    def things_to_diff(exp, act)
+      expect = mu_pp_for_diff exp
+      butwas = mu_pp_for_diff act
+      need_to_diff =
```
zenspider: use blank lines (paragraphs) to separate concepts

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
       @diff = o
     end
 
+    ##
+    # Returns things to diff [expect, butwas], or [nil, nil] if nothing to diff.
+    def things_to_diff(exp, act)
```
zenspider: no parens to match the project's style

lib/minitest/assertions.rb
```diff
@@ -50,27 +50,32 @@ def self.diff= o
       @diff = o
     end
 
+    ##
+    # Returns things to diff [expect, butwas], or [nil, nil] if nothing to diff.
```
zenspider: blank line between rdoc and method

zenspider:
I just did my first PR review via pure emacs... that was strange but neat... now to see if it made ANY sense on this end...

zenspider:
OK. Yeah. At least on the diff page. On the overview it looks backwards, but that's fine. I might file an issue with the emacs mode to reverse the order of submission if possible.

zenspider:
Overall, I like the idea even if I don't have a need for it. I can see its use case. I would do the logic slightly differently and change the rdoc accordingly... but I'm only really asking for some style changes. I can apply this with those already done.

lib/minitest/assertions.rb
BurdetteLamar: I think I've changed the things you requested.  Thanks, zenspider, for being patient.

BurdetteLamar:
Anything more I need to do here?

zenspider:
Tweaked this out a bit more and hand merged it in. I did some other work around diffing strings w/ mixes of newlines and escaped newlines and this collided. Sorry for the delay. Thank you!

BurdetteLamar:
Thanks, zenspider.

PR #797, opened 2019-08-04
Allow Minitest::Spec classes to be marshalled

barrettkingram:
Possible fix for https://github.com/seattlerb/minitest/issues/794

## Why
This PR takes the basic idea from https://github.com/blowmage/minitest-rails/pull/218 and applies it upstream to minitest. I think this problem is not necessarily specific to rails - it would be a problem with any parallel executor that uses DRb to synchronize parallel test runs across multiple processes. It feels like the fix belongs in minitest itself (since fixing it elsewhere requires monkey-patching minitest). Fixing it upstream also allows other gems like [minitest-spec-rails](https://github.com/metaskills/minitest-spec-rails) to take advantage of it.

## How
This change defines a unique constant for every spec class which will
allow them to be marshalled/unmarshalled and sent over the wire with DRb.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -268,6 +268,14 @@ class Minitest::Spec < Minitest::Test
         nuke_test_methods!
       end
 
+      cls_const = const_base = "Test__#{name.to_s.gsub(/\W/, '')}"
+      uniq_id = 2
+      while const_defined? cls_const
+        cls_const = const_base + uniq_id.to_s
+        uniq_id += 1
+      end
+      Minitest::Spec.const_set(cls_const, cls)
+
       children << cls
 
       cls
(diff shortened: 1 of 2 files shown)


blowmage:
I would like to see either of the solutions proposed in #716 re-evaluated along with this PR.

zenspider:
WTF? I swear I responded to this the same day I assigned it to myself... Unless there's a bunch of these all related.

@blowmage I've already done the `Result` proposal. Do you have any further input on why we need the _classes_ to be serializable if we've decoupled a result from the runnable?

blowmage:
The way I understand the Rails parallelism test setup, we want to reference the class on the parallel test process. But, if the class can't be serialized, like when it is an anonymous class generated by the spec dsl, then every use of the class will make a call to the originating process over DRb, and the test will effectively be run on the original process.

This is why I generated a (hopefully) unique but consistent constant for each spec class in minitest-rails, so each process can refer to the spec class by the constant and use it in each process.

I'm not 100% sure I understand the change you made, and I haven't tested it yet, so I don't yet know if it alone is sufficient.

zenspider:
I've got this working with a slightly different solution. Sorry it took me so long.

PR #798, opened 2019-08-14
support new Proc#to_s format

ko1:
We are discussing to change `Proc#to_s` notation from
  `#<Proc:0xXXXX@t.rb:2>`
to
  `#<Proc:0xXXXX t.rb:2>`
at https://bugs.ruby-lang.org/issues/16101

and minitest's test depends on this behavior.

This PR support both `@` or ` ` separation to path tests.
Comments for [Feature #16101] is also welcome.


zenspider:
I'm not sure the `@PATH` part should be merged as that stops minitest from normalizing the default `Object#inspect` output, no? Or, at the very least, it should be wrapped up in an optional.

zenspider:
I finally installed 2.7! Merged! Thank you!

PR #801, opened 2019-09-14
Tweak multithreading section of README

iHiD:
Two small tweaks to the README that I came across when reading it:

- Fixed header using which (I presume) was using markdown style rather than rdoc.
- Split out two sentences that didn't read easily when merged into one.


zenspider:
Good catch. Thank you. Done.

PR #802, opened 2019-09-21
Update a broken comments link about R^2

havenwood:
The old link was stale so I replaced it with the modern version of the same content.


zenspider:
Done. Thanks!

PR #804, opened 2019-09-24
Mention minitest-global_expectations in README

jeremyevans:
This library allows for backwards compatibility with older versions
of minitest, so you can continue to do `foo.must_equal 1` instead of
`_(foo).must_equal 1`.


zenspider:
Done! Thank you!

PR #805, opened 2019-09-27
Avoid using `match?` for supports old rubies

y-yagi:
This is breaking some libraries CI that supports old rubies (e.g. [Rails](https://buildkite.com/rails/rails/builds/63928#194ec831-37f0-4bc6-b702-c7965bdd061b/957-972)).


zenspider:
That's what? Ruby 2.2? How long with that go on? Any older ones?

zenspider:
Applied. I'll release soon.

PR #807, opened 2019-10-03
Only perform Minitest.autorun at_exits in initial process

jhawthorn:
It can be useful to `fork` within tests. We also use forking in Rails' test suite for test parallelization with process isolation.

Previously, when forking `Minitest.autorun`'s `at_exit` blocks would still be called, would run the after_run callbacks, and would change the exit status to 1.

This could be worked around using `Process.exit!` or often one could just ignore the status code of `fork`, but ideally forks could exit normally. I can't think of a case we'd want to run this code from a fork.

This commit changes `Minitest.autorun` to record the pid when it is called, and only runs the at_exit code in that process. I've added two (previously failing) tests to check we can exit with clean status codes.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -50,12 +50,15 @@ module Minitest
   # Registers Minitest to run at process exit
 
   def self.autorun
+    initial_pid = Process.pid
     at_exit {
+      next if Process.pid != initial_pid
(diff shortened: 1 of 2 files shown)


zenspider:
This is a dupe of #683. Closing but giving credit in the merge. Thank you

PR #813, opened 2019-10-25
use explicit Warning module

firien:
if an object has its own warn method, Deprecation warnings may have unexpected results.

An old legacy class of mine has warn defined and it accepts 0 arguments. This deprecation warning results in: ArgumentError: wrong number of arguments. Also, the actual depecration warning is not printed, so the results are, confusing to say the least.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -10,7 +10,7 @@ class Module # :nodoc:
       def #{new_name} *args
         where = Minitest.filter_backtrace(caller).first
         where = where.split(/:in /, 2).first # clean up noise
-        Warning.warn "DEPRECATED: global use of #{new_name} from #\{where}. Use _(obj).#{new_name} instead. This will fail in Minitest 6."
+        Kernel.warn "DEPRECATED: global use of #{new_name} from #\{where}. Use _(obj).#{new_name} instead. This will fail in Minitest 6."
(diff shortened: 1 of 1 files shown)


firien:
```ruby
require 'minitest/spec'
require 'minitest/autorun'

class Record
  def warn
    # do something
  end
  def active
    true
  end
end

describe Record do
  it "should be active" do
    Record.new.must_be :active
  end
end
```
```
Run options: --seed 31053

# Running:

E

Finished in 0.000853s, 1172.3328 runs/s, 0.0000 assertions/s.

  1) Error:
Record#test_0001_should be active:
ArgumentError: wrong number of arguments (given 1, expected 0)
    untitled 3:5:in `warn'

1 runs, 0 assertions, 0 failures, 1 errors, 0 skips
```

zenspider:
`Warning` doesn't exist on ruby 2.3, so I can't use this patch as-is... That said, the "correct" way to write specs now is this:

```ruby
describe Record do
  it "should be active" do
    _(Record.new).must_be :active # _ or value or expect
  end
end
```

which avoids this problem entirely.

firien:
> `Warning` doesn't exist on ruby 2.3

Roger that; what if we use `Kernel#warn` instead?

zenspider:
That works. Thank you!

Issue #814, opened 2019-11-02
New diff logic makes diffs for long multiline strings that contain '\\n' hard to use


mvz:
Because of the fix for #791, a multi-line string that contains `\\n` anywhere will no longer be diffed properly. For example:

```ruby
require 'minitest/autorun'

describe "foo" do
  it "bar" do
    foo = "very long text\nwith some\nescaped backslash\\nin it\nand more text"
    bar = "very long text\nwith some\nescabed backslash\\nin it\nand more text"
    assert_equal foo, bar
  end
end
```

Results in:
```
Failure:
foo#test_0001_bar [foo.rb:7]:
--- expected
+++ actual
@@ -1 +1 @@
-"very long text\nwith some\nescaped backslash\\nin it\nand more text"
+"very long text\nwith some\nescabed backslash\\nin it\nand more text"
```

This makes it hard to actually see where the difference is.

With minitest 5.11, I get the following, which is indeed also not quite right, but better for figuring out the difference:

```
  1) Failure:
foo#test_0001_bar [foo.rb:7]:
--- expected
+++ actual
@@ -1,5 +1,5 @@
 "very long text
 with some
-escaped backslash\
+escabed backslash\
 in it
 and more text"

```
```

zenspider:
Indeed... the problem was what to do when there are both escaped and unescaped backslash-n's in the string. The compromise around that was that if there are both, then there is probably a good reason for it (ie, a complex template sort of thing, like in your example) and the person doing that knows what they're doing, so don't confuse them further with the extra munging.

I'm not thrilled with it and I'm open to suggestions.

mvz:
I made a pull request with a suggestion.

Issue #791, opened 2019-05-07
Incorrect assert_equal output with escaped backslash before "n"


Ruin0x11:
When running this test:

```ruby
def test_escapes
  assert_equal "\\n", "\\"
end
```

I get this output:

```
 FAIL["test_escapes", #<Minitest::Reporters::Suite:0x000055748a415f88 @name="DecoderTest">, 0.0051554389065131545]
 test_escapes#DecoderTest (0.01s)
        --- expected
        +++ actual
        @@ -1,2 +1 @@
        -"\
        -"
        +"\\"
        test/unit/decoder_test.rb:443:in `test_escapes'
```

Shouldn't the expected output print `\\n` instead of an actual newline instead?

zenspider:
This is a wonderfully fucked-up edge case. Thank you. It is thoroughly flummoxing me.

I think my current strategy is to extend the pretty diff with extra logic to make sure that no string has both real newlines and escaped ones... but I need to translate the above into the edge case combos I'm thinking of. Debugging diffed diffs is a bitch tho.

PR #821, opened 2019-12-06
Only unescape real newlines for diff

mvz:
This is a fix for #814 that should make the #791 case look good as well. The basic idea is to *only* unescape real newlines, and leave everything else as-is. I think this is the method that gives the least surprising results.


akostadinov:
Any updates?

Dustin4444 (approved):
Fixes #1045 @

Issue #819, opened 2019-11-24
Add example for #_(value) method with block to Expectations module


stomar:
The docs for Minitest::Expectations only show examples for the value wrapper used with non-block values, and the examples for the expectation methods only show the old syntax using `proc { ... }`.

An example for using the value wrapper with a block can be found in the docs for Minitest::Spec::DSL::InstanceMethods:

```ruby
_ { 1 + "1" }.must_raise TypeError
```

but it's not very discoverable.

It might be rather helpful to add this example to the module docs for Minitest::Expectations.

PR #826, opened 2020-01-11
Add example for value wrapper with block to Expectations module

stomar:


stomar:
Closes #819.

zenspider:
Done. Thanks!

PR #828, opened 2020-01-29
Add helpful section upgrading to MT6

fzakaria:
Add section explaining use of [ffast](https://github.com/jonatas/fast/) to help migrate code easily to the new expectation sequence.

I recently migrated our version of Minitest on a very large codebase and we've been hit with a **ton** of _DEPRECATED_ warnings.

Relying on AST transformation is a great way to easily migrate the codebase to use the new style.

There are even more advanced concepts in the _fast_ library such as "experiments"; which can run minitest following every change. The amount documented here should be a great starting point for others thought!

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -376,6 +376,34 @@ Using our example above, here is how we might implement MyCI:
 
 == FAQ
 
+== Easing the transition to Minitest 6
+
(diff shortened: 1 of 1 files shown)


zenspider:
No thank you...

PR #830, opened 2020-01-30
Add ability to silence warnings

### LINKED ISSUE #829: Use assert_nil if expecting nil from -- No way to solve for dynamic asserts
We have some tests that do something similar to:

```
it 'some test'
    actual = run_task()
    expected = {
        :a  => "a,
        :b => nil,
        :c => "c"',
        :d => nil,
        :e => "e",
        :f => "f",
        :g => "g",
    }
    expected.each{|k, v| expect(actual[k]).must_equal v, k}
end
```
It's a simply way to make a single line assertion handle a Hash of potential values.
Unfortunately, Minitest is warning with:

```DEPRECATED: Use assert_nil if expecting nil from $FILE. This will fail in Minitest 6.```

Can this be removed -- I don't see why the forcing of `assert_nil` is required here. 
I understand the deprecation notice for the expectation monad.

Furthermore, can the **warn** be toggled off via environment variable?

fzakaria:
Add an environment variable (_MT_SILENCE_) to silence the **DEPRECATED** warnings.

Although they are helpful, in a large codebase, they can pose a lot of noise for those that are not tasked with migrating the codebase. I am slowly migrating our codebase however our developers are frustrated with the number of warnings in by Minitest in the logs.

I would very much appreciate this happy medium.

fixes #829

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -224,7 +224,7 @@ module Minitest
           where = Minitest.filter_backtrace(caller).first
           where = where.split(/:in /, 2).first # clean up noise
 
-          warn "DEPRECATED: Use assert_nil if expecting nil from #{where}. This will fail in Minitest 6."
+          warn "DEPRECATED: Use assert_nil if expecting nil from #{where}. This will fail in Minitest 6." unless ENV["MT_SILENCE"]
         end
       end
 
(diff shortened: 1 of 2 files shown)


fzakaria:
I've linked to #829 as well as this will fix the issue of seeing the warnings however there is a larger discussion about the example in the issue. The warning is counter-productive to writing a nice terse assertion for a composite object.

zenspider:
`-W0` silences warnings (as does `$-W = nil` in your test helper.

PR #833, opened 2020-02-26
Reseed RNG before each test

casperisfine:
This ensure that for a given `--seed` value, a test will always start from the same seed.

Otherwise previous tests using the RNG, would leak state.

This is something we had for a very long time as a global `setup`, I figured it made sense upstream.

A recent commit to ruby reminded me of it: https://github.com/ruby/ruby/commit/7c1553e91db07fed4fed3287b50304f1c69a685d

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -323,6 +323,7 @@ module Minitest
 
       with_info_handler reporter do
         filtered_methods.each do |method_name|
+          Random.srand options[:seed] if options[:seed]
           run_one_method self, method_name, reporter
         end
       end


zenspider:
I dunno about this one. I suspect it'll break a lot of tests out there.

I have `srand`ed before shuffling classes and methods for each runnable so that will stabilize that aspect (previously user/test use of srand could throw run order off)... 

without more analysis to show that this won't "break the world" I think I'd rather leave it up to the user/test to do it themselves. I've added `Minitest.seed` in order to make this more readily available.

zenspider:
I agree that could eliminate certain types of errors, but it could introduce others as well. By making the seed available the user can opt in where applicable:
```ruby
def setup
  super
  srand Minitest.seed
end
```

PR #834, opened 2020-02-29
Return true on a successful refutal

jusleg:
Updates the `refute` method to make it match its lexical meaning. A
successful refutal will return `true` and a failed one will raise. Given
that refute is used by other assertion definitions, the following
methods now return `true` on successful evaluation:

- path_wont_exist
- wont_include
- wont_be
- wont_be_close_to
- wont_be_within_epsilon
- wont_equal
- wont_be_instance_of
- wont_be_kind_of
- wont_match
- wont_be_empty
- wont_be_same_as
- wont_be_nil
- wont_respond_to

Having it return `true` means that you could do something like

```ruby
assert(refute(thing.something?))
```

Although I'm not sure why one would do that

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -628,7 +628,7 @@ module Minitest
 
     def refute test, msg = nil
       msg ||= message { "Expected #{mu_pp(test)} to not be truthy" }
-      not assert !test, msg
+      assert !test, msg
     end
 
     ##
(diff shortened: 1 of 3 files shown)


zenspider:
> Although I'm not sure why one would do that

to test minitest. :P

zenspider:
NICE. Thank you. It also nudged me to refactor all of those `.must_equal true` lines into a custom assertion.

Done. Thank you again.

PR #824, opened 2019-12-31
If a callable mock includes kwargs, Ruby 2.7 warns

kddeisz:
This commit explicitly forwards kwargs to remove the warning, and adds a test to assert that it was silent.

PR #845, opened 2020-04-09
update mock.rb stub to properly call keyword opts

SampsonCrowley:
ruby 2.7 introduces deprecations for positional keyword arguments. this fixes them

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -209,7 +209,7 @@ class Object
   #     end
   #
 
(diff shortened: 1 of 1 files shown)


yahonda:
It looks like duplicate of #824.

zenspider:
Thanks for the contribution! Unfortunately I can't accept this.

This doesn't have any tests and is missing some edge cases. I think I've got all of this covered now, so I'm going to close this. I'll release my changes soonish.

PR #862, opened 2020-12-26
Bump Ruby version to include 3.1 (ruby-head)

### LINKED ISSUE #861: ruby-head compatibility
We're testing `rspec-expectations` for compatibility with `minitest`, and we run it on `ruby-head` as well. This [fails](https://github.com/rspec/rspec-expectations/pull/1245/checks?check_run_id=1611861008) due to no `minitest` version being compatible with Ruby 3.1 https://github.com/seattlerb/minitest/blob/dc6ad179ceac3be92375ac1236f0ee19b9f09df4/Rakefile#L14

Do you think it would be possible to relax the Ruby version constraint at some point?

pirj:
3.0 was released, and ruby-head is now 3.1
See https://github.com/rspec/rspec-expectations/pull/1245/checks?check_run_id=1611861008

Fixes #861

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,7 @@ Hoe.spec "minitest" do
 
   license "MIT"
 
-  require_ruby_version [">= 2.2", "< 3.1"]
+  require_ruby_version [">= 2.2", "< 3.2"]
 end
 
 desc "Find missing expectations"


MSP-Greg:
This breaks quite a few repo's CI.  A new release would be appreciated...

koic:
@zenspider Thank you for your maintenance. Some repositories are requiring for this change. I hope it will be released at your convenience.

Rakefile
```diff
@@ -11,7 +11,7 @@ Hoe.spec "minitest" do
 
   license "MIT"
 
-  require_ruby_version [">= 2.2", "< 3.1"]
+  require_ruby_version [">= 2.2", "< 3.2"]
```
casperisfine, line 14: IMHO it would be better to simply get rid of the upper bound. There's no reason to assume future version will break minitest, and this kind of restriction make testing ruby-head a huge pain.

tenderlove:
I'm running in to this issue as well.  Here is my current workaround:

```
$ cat x.rb
module Gem
  class Specification
    def required_ruby_version
      Gem::Requirement.default
    end
  end
end
$ RUBYOPT="-I"(pwd)" -rx" bundle
```

I'm using Fish shell, so the final command might be different for other folks. But the point is to monkeypatch `required_ruby_version`

Rakefile
MSP-Greg: I haven't looked to see if Ruby is following semantic versioning, but I believe that several several std-lib gems are.

Hence, since Minitest is working with 3.0, the following would certainly be appropriate:
```ruby
require_ruby_version [">= 2.2", "< 4.0"]
```

zenspider:
Fixed and released! Thanks everyone!

PR #865, opened 2021-01-12
Limit spec dsl defined classes to own test methods

amckinnie:
**Issue:**
Currently, minitest spec styles deals with tests being included from preceding or ancestor classes within the DSL with the method `Minitest::Spec::DSL#nuke_test_methods!`. This will wipe out existing test methods that have been defined at the time of class initialization (`Minitest::Spec::DSL#create`). However, when tests are defined after a block is declared, they are still included when the instance methods for the class are pulled (since they're pulled with all the ancestor instance methods.

In this example:
```
class MyTest < Minitest::Test
  describe "foo" do end
  def test_name; end
end
```
`test_name` will be run twice, once as part of class `MyTest` and once as part of class `MyTest::foo`. This will occur for each describe block (or alias), so if there are multiple describe blocks in a test file, it will run for each of them.

**Proposed Solution:**
For classes defined by the spec DSL, limit test methods pulled to that of the class itself.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -266,6 +266,10 @@ class Minitest::Spec < Minitest::Test
         @desc = desc
 
         nuke_test_methods!
+
+        def self.methods_matching re
+          public_instance_methods(false).grep(re).map(&:to_s)
+        end
       end
 
       children << cls
--- a/test/minitest/test_minitest_spec.rb
+++ b/test/minitest/test_minitest_spec.rb
@@ -989,6 +989,18 @@ class TestMeta < MetaMetaMetaTestCase
     assert_respond_to y.new(nil), "xyz"
     assert_respond_to z.new(nil), "xyz"
   end
+
+  def test_methods_around_describe
+    y = nil
+    x = describe "top-level thingy" do
+      def test_before_desc; end
+      y = describe "second thingy" do end
+      def test_after_desc; end
+    end
+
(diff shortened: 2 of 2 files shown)


lib/minitest/spec.rb
```diff
@@ -266,6 +266,10 @@ def create name, desc # :nodoc:
         @desc = desc
 
         nuke_test_methods!
+
+        def self.methods_matching re
+          public_instance_methods(false).grep(re).map(&:to_s)
+        end
```
zenspider, line 272: This doesn't really make sense here, does it? Maybe I'm confused.

zenspider (changes requested):

lib/minitest/spec.rb
amckinnie: The class created by `Class.new` will inherit from Minitest:Spec or some other subclass (Minitest::Spec#spec_type). `Class.new` allows you to run code or create methods for that dynamic subclass which will be defined with the call to describe.

I had initially approached it by changing the method in Runnable, though that would cause issues with inheritance (since we wouldn't want to upset the current expected behavior of a test class inheriting from another running the tests of the inherited class). Thus, we'd need only the classes defined by `describe` to have this behavior, which is why we'd need this here.

zenspider:
Why not in Spec?

amckinnie:
> Why not in Spec?
We could, though with the inheritance issue still applies.
```
class ATest < Minitest::Spec
  ...
end

class BTest < ATest
  ...
end
```
This would currently run everything from ATest in BTest. Would we no longer want to do that?

PR #867, opened 2021-01-16
Expose command line options via Minitest.option hash

Ryan1729:
This is an adjustment of PR https://github.com/seattlerb/minitest/pull/657 to use `option`, singular and opposed to `options` plural, as suggested in [this comment](https://github.com/seattlerb/minitest/pull/657#discussion_r88561532).

I did this because I would also like to be able to make all test randomness depend only on the passed seed value. 

https://github.com/seattlerb/minitest/pull/833 is also potentially relevant.

Closes #657

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -459,9 +459,9 @@ able to require minitest and run your tests.
 === How can I access the minitest options (for example the seed value) in my tests?
 
 If you like to use for example the generated or given seed value in your tests
-you can get the minitest options now with +Minitest.options+ from everywhere.
+you can get the minitest options now with +Minitest.option+ from everywhere.
 This is usefull to feed the random seed generator before testing something with random input,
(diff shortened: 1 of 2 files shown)


lzap:
I would really love to have this. I want to hardcode `--verbose` option in my projects.

Qqwy:
I would also love this feature.

Why was the change from `options` -> `option` (singular) made in the last commit?

Ryan1729:
@Qqwy 
> Why was the change from `options` -> `option` (singular) made in the last commit?

I made that change because singular `option` was asked for in [this 2016 comment](https://github.com/minitest/minitest/pull/657#discussion_r88561532) on the original PR. That request was reaffirmed in [this 2019 comment](https://github.com/minitest/minitest/pull/657#discussion_r292366329), so it seemed like making that change might increase the chances of this getting merged.

zenspider:

PR #871, opened 2021-02-09
Enable Ruby deprecation warnings by default

casperisfine:
Ruby 2.7.2 turned deprecation warnings off by default, and it is now recommended that test frameworks do enable them.

See https://bugs.ruby-lang.org/issues/17591 for more context.


zenspider:
How strange. Thanks! Added and will be released shortly-ish.

Issue #872, opened 2021-03-12
parallel mode, but with ractors


HoneyryderChuck:
I've been looking at the feasibility of adding a runner, akin to Parallel, where workers are ractors. Although I expect most test suites will break under it, it could help the community move towards "ractor-friendly" components.

The steps towards it seem to be:

* [Remove default mutex from reporters (only needed in parallel mode, should be defined there)](https://github.com/seattlerb/minitest/blob/3c6576a51f4e266996e3459d7a0dd054eb4c87f7/lib/minitest.rb#L555)
* Build ractor-friendly "queue" (currently, can be done using a ractor "pipe")

Does this seem to be a good addition? Is anyone already working on something? Let me kow what you think before start doing anything.

zenspider:
Nobody I know is doing such a thing... I honestly have no idea if it'd be a useful thing or not.

Happy to do refactorings that make sense if they help.

HoneyryderChuck:
Probably not that useful at this point, as a lot of code is known to break under ractors. Still, it would be cool to lay the groundwork for when there are less limitations.

I'll see if I can think about something. I had a plan when I wrote the message, but it's been some time, so I'll have to revisit the strategy.

zenspider:
I'm gonna close this. I think it is out of scope for minitest at this point. I'll poke at your PR tho.

HoneyryderChuck:
No problem. Just letting you know that I paused work on it until [this bug gets fixed](https://bugs.ruby-lang.org/issues/17998). Still some work onn ractors til they're reliable.

PR #878, opened 2021-06-14
wip: moved setup of reporter mutex to the parallel module

HoneyryderChuck:
Step towards #872 

Removes report synchronization code into the `parallel` module.


zenspider:
Given that your ractor bug is still open, preventing this from moving forward, I'm gonna close this for now. Happy to revisit in the future if it becomes a possibility.

PR #879, opened 2021-06-17
Close then unlink tempfiles on Windows

nobu:
In ruby/ruby test actions, number of "leaked tempfile" messages are shown on Windows.

As Windows disallows removing open files, `Tempfile#unlink` fails silently before `#close`.
Close then unlink by `#close!` instead.

https://github.com/rubygems/rubygems/pull/4681


zenspider:
Done! Thank you! Sorry that took so long.

PR #882, opened 2021-07-20
Use Etc.nprocessors by default in order to maximize cpu usage.

tonytonyjan:
2 threads by default is too few in 2021. A more modern way is to detect the number of CPU cores automatically.


zenspider:
Fantastic. Thank you. Should be released soonish.

tonytonyjan:
@zenspider Thank you!

PR #885, opened 2021-11-25
Don't pass implicit blocks for method_missing stubs

danielma:
Switch from `send :define_method` to `class_eval` for stubbed methods
that would otherwise by handled by `method_missing`

With the previous implementation, the block passed to `stub` is in
scope during `define_method`, which means that once a method on an
object has been stubbed, subsequent method calls will always receive
that implicit block.

I've added a new test that presents this behavior with the existing
implementation, and works properly with this change.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -217,9 +217,11 @@ class Object
     metaclass = class << self; self; end
 
     if respond_to? name and not methods.map(&:to_s).include? name.to_s then
-      metaclass.send :define_method, name do |*args|
-        super(*args)
-      end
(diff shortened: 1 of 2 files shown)


zenspider:
Why does this matter?

danielma:
It matters because the existing implementation can cause behavior changes to the original class. This change attempts to improve stub cleanup to be more comprehensive

zenspider:
Apparently I can only transfer issues to another project, not a PR, even if the patch is compatible. If you're still interested in this, please move this PR to https://github.com/minitest/minitest-mock. Thank you.

Closing this.

PR #886, opened 2021-12-06
Add GitHub Actions CI, small platform lib & test fixes

MSP-Greg:
GitHub Actions (GHA) has limits to concurrent jobs. The macOS limit is 5 for Free/Pro/Team accounts, and the combined Ubuntu/Windows limit is 20 for Free accounts. So, tried to select the jobs to run based on that. CI often completes in ~ 2 minutes.

Tried to add JRuby and TruffleRuby, both seem to have intermittent issues. Hopefully I can do some more work with it. Many other CI systems have a similar 'allow-failure' syntax for jobs. Actions does not. We could add another workflow for JRuby and TruffleRuby that won't fail the commit/PR, but would log failures/errors so someone can fix the issues?

I added TESTOPTS=-v, which can be removed.

Misc Notes

The standard 'action' to load Ruby is `ruby/setup-ruby`.  Currently, it is not running the current Windows Image, it also cannot work with Windows 'UCRT' builds. 'UCRT' is a new gcc based tool chain that will be the chain used when Windows Ruby 3.1 is released.  So, for the time being, `MSP-Greg/ruby-setup-ruby@win-ucrt-1` is being used, which wraps `ruby/setup-ruby`.   

Secondly, in the past, some other CI systems' Ruby trunk/head builds were just cron jobs that built Ruby, but did not run any of its test suites.  All of the trunk/head builds available with GHA are fully tested and used extensively.  JFYI, if a cron trunk/head build does not pass the test suites, it does not replace the previous passing build.

Finally, in the past, repos without Actions CI will not run the CI when it's added in a PR. We'll see what happens with this... You can look at a run in my fork: https://github.com/MSP-Greg/minitest/actions/runs/1542633084

Closes #683

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,88 @@
+name: CI
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  test:
+    name: >-
+      test-${{ matrix.os }} ${{ matrix.ruby }}
+    env:
+      TESTOPTS: -v
+
+    runs-on: ${{ matrix.os }}
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-20.04, ubuntu-18.04, macos-10.15, macos-11, windows-2022 ]
+        ruby: [ 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, '3.0', head ]
+        include:
+          - { os: windows-2022, ruby: mingw }
+          - { os: windows-2022, ruby: ucrt  }
+
+        exclude:
+          - { os: windows-2022, ruby: head  }
+          - { os: macos-10.15 , ruby: 2.6   }
+          - { os: macos-10.15 , ruby: 2.7   }
+          - { os: macos-10.15 , ruby: '3.0' }
+          - { os: macos-10.15 , ruby: head  }
+          - { os: macos-11    , ruby: 2.2   }
+          - { os: macos-11    , ruby: 2.3   }
+          - { os: macos-11    , ruby: 2.4   }
+          - { os: macos-11    , ruby: 2.5   }
+
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v2
+
+      - name: load ruby
+        uses: MSP-Greg/ruby-setup-ruby@win-ucrt-1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: install hoe
+        run:  gem install hoe
+
+      - name: rake test
+        run:  |
+          rake test
+          ruby -v
+        timeout-minutes: 3
+
+  isolated:
+    name: >-
+      isolated-${{ matrix.os }} ${{ matrix.ruby }}
+    env:
+      TESTOPTS: -v
+
+    runs-on: ${{ matrix.os }}
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-20.04, windows-2022 ]
+        ruby: [ 2.3, '3.0' ]
+        include:
+          - { os: macos-10.15 , ruby: 2.3   }
+          - { os: macos-11    , ruby: '3.0' }
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v2
+
+      - name: load ruby
+        uses: MSP-Greg/ruby-setup-ruby@win-ucrt-1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: install hoe
+        run:  gem install hoe
+
+      - name: rake test:isolated
+        run:  |
+          rake test:isolated
+          ruby -v
+        timeout-minutes: 3
--- a/test/minitest/test_minitest_assertions.rb
+++ b/test/minitest/test_minitest_assertions.rb
@@ -28,6 +28,9 @@ class TestMinitestAssertions < Minitest::Test
 
   RUBY18 = !defined? Encoding
 
+  # not included in JRuby
+  RE_LEVELS = /\(\d+ levels\) /
+
   class DummyTest
     include Minitest::Assertions
     # include Minitest::Reportable # TODO: why do I really need this?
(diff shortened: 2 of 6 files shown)


test/minitest/test_minitest_reporter.rb
```diff
@@ -13,6 +13,10 @@ class TestMinitestReporter < MetaMetaMetaTestCase
   attr_accessor :r, :io
 
   def new_composite_reporter
+    # Ruby bug in older versions of 2.2 & 2.3 on all platforms
+    # Latest Windows builds were 2.2.6 and 2.3.3.  Latest Ruby releases were
+    # 2.2.10 and 2.3.8.
+    skip if /mingw|mswin/ =~ RUBY_PLATFORM && RUBY_VERSION < '2.4'
```
zenspider, line 19: Same with `windows?` 

https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L947-L1010 few others too

test/minitest/test_minitest_test.rb
```diff
@@ -848,19 +848,25 @@ def non_verbose
     $VERBOSE = orig_verbose
   end
 
+  def sample_test_case(rand = nil)
+    srand(rand) if rand
+    Class.new FakeNamedTest do
+      100.times do |i|
+        define_method("test_#{i}".to_sym) { assert true }
```
zenspider, line 855: you never need to `to_sym` for `define_method`

test/minitest/test_minitest_test.rb
```diff
@@ -848,19 +848,25 @@ def non_verbose
     $VERBOSE = orig_verbose
   end
 
+  def sample_test_case(rand = nil)
+    srand(rand) if rand
```
zenspider, line 852: no parens on the srand above please

why is `rand` optional if you use it in 100% of the calls?

test/minitest/test_minitest_assertions.rb
```diff
@@ -1366,7 +1371,7 @@ def test_skip_until
     d0 = Time.now
     d1 = d0 + 86_400 # I am an idiot
 
-    assert_output "", /Stale skip_until \"not yet\" at .*?:\d+$/ do
+    assert_output "", /Stale skip_until \"not yet\" at #{__FILE__}:\d+$/ do
```
zenspider, line 1374: I rather dislike absolute paths and am thinking of changing it... Can you revert this bit or does it serve a purpose I don't see yet?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,88 @@
+name: CI
+
+on: [push, pull_request, workflow_dispatch]
```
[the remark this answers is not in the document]
zenspider: Damnit... got bit by `pending`... I wrote this days ago :P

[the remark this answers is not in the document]
MSP-Greg: This is messy, and I might say non-intuitive...

1. A main issue with the `on:` syntax is how do the maintainers submit PR's, and is it a concern for people to be able to run CI in their fork.  The following mean that only the 'main' branch will run CI.  If I would like to do a PR but run CI in my fork (for testing, checking platforms/os's that I can't run locally, etc), the only way to do so is by using the 'main' branch, which many people would prefer not to do.  I'd make a branch for the PR in my fork, and leave main alone.
```yaml
    branches:
      - main
```

But, as to maintainers, if you create a branch here, and use it for a PR, and the above is not included, then the PR will run both as a PR and as a branch.  Does that make sense? 

2. `workflow_dispatch` - this allows maintainers to run a workflow on any branch.  Sometimes one may have CI failures that are not related to anything in the repo, maybe a network hang, etc.  You can re-run workflows, but that overwrites the 'bad' run.  `workflow_dispatch` allows one to re-run it without overwriting.

3. `pull_request` - as I tried to mention, if a repo does not have any workflows, a PR with workflows will not run.  Once a repo is using Actions, PR's can be submitted that change the workflows,

test/minitest/test_minitest_assertions.rb
MSP-Greg: The RegEx won't work with windows paths.  Example `C:\Windows`.  Sorry, I thought I added a comment somewhere...

test/minitest/test_minitest_test.rb
MSP-Greg: Don't recall.  I might have called it without a parameter while writing/testing it.  I'll remove both.

.github/workflows/ci.yml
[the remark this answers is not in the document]
zenspider: 1. Very unintuitive... yes.

2. Mmmm... We don't use `workflow_dispatch` at work and GHA _definitely_ provides access to previous builds even with a manual rerun. They've even recently added some UI to be able to go to the previous.

3. that makes sense. thank you.

test/minitest/test_minitest_assertions.rb
zenspider: it _should_ work with windows paths... that `$` should force the regexp to keep trying after matching `c:\`, no? We can remove the `?` if that's what's interfering but I didn't think it was.

```
>> re = /Stale skip_until \"not yet\" at .*?:\d+$/
=> /Stale skip_until \"not yet\" at .*?:\d+$/
>> s = "Stale skip_until \"not yet\" at c:\\blah.rb:42"
=> "Stale skip_until \"not yet\" at c:\\blah.rb:42"
>> s =~ re
=> 0
```

.github/workflows/ci.yml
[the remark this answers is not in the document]
MSP-Greg: `workflow_dispatch` doesn't in itself do anything.  It allows maintainers/owners to run a workflow file. Maybe a dependency was updated and there were issues in another repo, and one wants to run CI without a PR or commit...

> They've even recently added some UI to be able to go to the previous.

I haven't done a rerun for a while, I hadn't seen that.

test/minitest/test_minitest_test.rb
MSP-Greg: I'll try, but Ruby's doc seems to only show a symbol as a parameter...

MSP-Greg: You're correct.  Fixed.

test/minitest/test_minitest_assertions.rb
MSP-Greg: Sorry. I think I figured out what I did.  The rexgexp works, I tried the change assuming the test was bad.  Actually, the code was bad, so it was fixed in `assertions.rb` in the first commit.  I accidentally left the test change in.

It's only an issue with Windows, and although `skip_until` makes sense, I don't recall seeing it in use.

Anyway, removed the change.

MSP-Greg:
Pushed all the changes, see: https://github.com/MSP-Greg/minitest/actions/runs/1568308489

zenspider:
Done. Thank you!

PR #887, opened 2021-12-06
Update deprecation message for block expectations

blowmage:
Update block expectations' deprecation message to use the block syntax.

[refs seattlerb/minitest#856]


zenspider:
Done. Thanks!

PR #888, opened 2021-12-10
`minitest-heat` Reporter

garrettdimon:
Adding [minitest-heat reporter](https://github.com/garrettdimon/minitest-heat/wiki) to the list.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -647,6 +647,7 @@ minitest-growl              :: Test notifier for minitest via growl.
 minitest-happy              :: GLOBALLY ACTIVATE MINITEST PRIDE! RAWR!
 minitest-have_tag           :: Adds Minitest assertions to test for the existence of
                                HTML tags, including contents, within a provided string.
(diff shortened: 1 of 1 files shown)


zenspider:
This looks awesome!

One request (for your code)... please put some sort of header/footer on it so it can be found... Our CI at work runs on ~40 jobs across hundreds of test files and I'll need to grab the logs and combine all the results to make it worthwhile.

zenspider:
Done. Thanks!

garrettdimon:
I think I know what you mean about the header/footer, but do you have an example from other reporters so I can make sure it's a good solution? And I assume it's something that would be fine being configurable so teams/contexts that don't need it don't have the extra visual noise?

zenspider:
Something like that… No examples off the top of my head. honestly, even if you just do a stupid markdownish header (eg `## Heat Map`) and had a consistent format after that, I’m sure it’d be extractable enough.

I guess the question remaining is: how to extract multiple and combine them in a meaningful way?

PR #811, opened 2019-10-23
[WIP] Fix unstubbing of not defined methods

GermanDZ:
This is only a POC breaking the tests to establish the existence or not of a bug.

Summary: After stubbing a method not really defined, but the module (probably will behave the same for classes) accepts a message (respond to...) and it's managed by meta-prorgramming used "method_missing" then the unstubbing is not working.

PR #891, opened 2022-01-17
Fixing an issue with a spoiling of the object singleton classes

alekseyl:
.stub method defined in minitest/mock executing an alias_method against singleton_class under the hood.  
Running an alias_method over a singleton_class without running a 'remove_method' later will keep singleton_class 'dirty' and running undef_method might clear more than needed.  

Usually this will not lead to any consequences, but if you are stubbing methods of a global objects and trying to enrich functionality of their ancestors chain 'after' the the first stubbing is done, then you might fall into trouble.

Even though the probability for such a case is tiny, it's not 0 and singleton class should be kept clean.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -216,7 +216,9 @@ class Object
 
     metaclass = class << self; self; end
 
-    if respond_to? name and not methods.map(&:to_s).include? name.to_s then
+    if respond_to?( name ) && !methods.map(&:to_s).include?( name.to_s )
+      # this mean that 'name' is a public method which is expected
+      # to be defined dynamically somewhere inside ancestors chain
       metaclass.send :define_method, name do |*args|
         super(*args)
       end
@@ -237,8 +239,7 @@ class Object
 
     yield self
   ensure
-    metaclass.send :undef_method, name
-    metaclass.send :alias_method, name, new_name
-    metaclass.send :undef_method, new_name
+    metaclass.send :remove_method, name
+    metaclass.send :remove_method, new_name
   end
 end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -213,10 +213,11 @@ class Object
 
   def stub name, val_or_callable, *block_args
     new_name = "__minitest_stub__#{name}"
+    singleton_has_stubbing_method = singleton_methods.map(&:to_s).include?( name.to_s )
 
     metaclass = class << self; self; end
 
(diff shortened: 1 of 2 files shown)


luke-gru (commented):
What about when you're stubbing a singleton class method instead of an instance method? This would erase the method on the singleton class, unless I'm reading it wrong.

GermanDZ:
> What about when you're stubbing a singleton class method instead of an instance method? This would erase the method on the singleton class, unless I'm reading it wrong.

Also consider these other cases: https://github.com/minitest/minitest/pull/811

alekseyl:
> What about when you're stubbing a singleton class method instead of an instance method? This would erase the method on the singleton class, unless I'm reading it wrong.

hm, what about the testcase in the PR:

```
def test_class_singleton_class_will_not_loose_its_method
    _class = Class.new do
      def self.clean; :dirty end
    end

    _class.stub(:clean, :clean) do |s|
      assert_equal( s.clean, :clean )
    end
    # method is still present and unchanged
    assert_equal( _class.clean, :dirty )
  end
```
Looks like its covering your concern

mattrobinson-shopify:
I encountered this issue in a rails codebase

Here's a minimal reproduction of what I was doing

```ruby
# minitest_test.rb
require "test_helper"

class MinitestTest < ActiveSupport::TestCase
  use_memcached

  test "can expect write after minitest stub" do
    Rails.cache.stub(:write, nil) do
      assert true
    end
    ActiveSupport::Cache::ProxyCacheStore.any_instance.expects(:write).with("foo", "bar")
    Rails.cache.write("foo", "bar")
  end
end
```

And the output of running this

```
% ruby -Ilib:test minitest_test.rb
* Loading test environment
Started with run options --seed 54969

FAIL
MinitestTest#test_can_expect_write_after_minitest_stub
    not all expectations were satisfied
    unsatisfied expectations:
    - expected exactly once, invoked never: #<AnyInstance:ActiveSupport::Cache::ProxyCacheStore>.write("foo", "bar")

1/1: [===============] 100% Time: 00:00:00, Time: 00:00:00

Finished in 0.64460s
1 tests, 2 assertions, 1 failures, 0 errors, 0 skips
```

If remove the `Rails.cache.stub(:write, nil) do` block or apply the changes suggested in this PR, the test passes.

In the meantime, I've just replaced the `Rails.cache.stub(:write, nil) do` with mocha's `Rails.cache.expects(:write)` as it's a stronger assertion anyway, but figured I'd chime in here to let minitest know this is still an issue.

PR #893, opened 2022-01-22
Replace http with https whenever such link exists

apatniv:
Also [about section on the right side of github page](https://github.com/seattlerb/minitest) uses a http instead of https. I can't fix it. 

Thank you


zenspider:
done. thanks!

PR #897, opened 2022-01-29
+1 minitest extesion gem called 'stubberry'

alekseyl:
Introducing stub_must, assert_method_called, + ActiveRecord stubbing object inside the block by given id and so.
  
https://github.com/alekseyl/stubberry


zenspider:
Done! Thanks!

PR #899, opened 2022-02-18
Update the test suite to handle ASCII-8BIT renaming.

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/18576
Ref: https://github.com/ruby/ruby/pull/5571

I'd like to rename `ASCII-8BIT` into `BINARY` for Ruby 3.2 (basically invert the `BINARY -> ASCII-8BIT` alias), and since minitest's test suite is integrated into Ruby it breaks the build.


zenspider:
If you still have the alias (just reversed), then how/why is it broken?

zenspider:
OK. This is approved and "merged"... will prolly release sometime this weekend once I triage some more.

zenspider:
P.S. I left some feedback on your PR for the C side

PR #902, opened 2022-02-22
Add stub test for public module method that was private

gmcgibbon:
Adds a test to stub a method that was previously private. For some reason, this fails on Ruby 3.1.0. This is a regression in CRuby somewhere, probably related to method aliasing. Still, I believe the test is worth adding to prevent future regressions.


zenspider:
That's a really strange good find... glad it got fixed in ruby. I'll see about guarding this in a way that passes on the guilty versions and notifies once there's a 3.1 that works.

zenspider:
Just rebased to force a retest... still brokey... on 3.1 and head... Odd because you can see that bug getting fixed 7 months ago and I could have sworn there's been a patch release since then?

gmcgibbon:
https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-1-2-released/

It looks like 3.1.2 was released after this was fixed in https://github.com/ruby/ruby/pull/5592. CI is using 3.1.2, so I guess it wasn't included.

zenspider:
Better go dig up a CVE worthy bug so they have a forcing function to do another release. 😛

zenspider:
Apparently I can only transfer issues to another project, not a PR, even if the patch is compatible. If you're still interested in this, please move this PR to https://github.com/minitest/minitest-mock. Thank you.

Closing this.

PR #903, opened 2022-02-26
Update Actions - Ruby 3.1, switch to ruby/setup-ruby

MSP-Greg:


zenspider:
I think this is fully covered in the current CI ... can you sanity check it for me?

PR #904, opened 2022-03-26
Support keyword args in mocks and ruby 3

benortiz:
Without changing the API for mock args, falling back to the ruby 2
compatibility mode seems like the easiest way to continue to support
keyword arguments in mocks and ruby 3.

### DIFF
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -169,6 +169,8 @@ module Minitest # :nodoc:
       retval
     end
 
+    ruby2_keywords :method_missing if respond_to?(:ruby2_keywords, true)
+
     def respond_to? sym, include_private = false # :nodoc:
       return true if @expected_calls.key? sym.to_sym
       return true if @delegator && @delegator.respond_to?(sym, include_private)
--- a/test/minitest/test_minitest_mock.rb
+++ b/test/minitest/test_minitest_mock.rb
@@ -275,6 +275,18 @@ class TestMinitestMock < Minitest::Test
     assert_mock mock
   end
 
+  def test_mock_block_is_passed_keyword_args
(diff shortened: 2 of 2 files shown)


test/minitest/test_minitest_mock.rb
```diff
@@ -275,6 +275,18 @@ def test_mock_block_is_passed_function_params
...
+    arg1, arg2, arg3 = :bar, [1, 2, 3], { :a => "a" }
+    mock = Minitest::Mock.new
+    mock.expect :foo, nil do |k1:, k2:, k3:|
+      k1 == arg1 && k2 == arg2 && k3 == arg3
+    end
+
+    mock.foo(k1: arg1, k2: arg2, k3: arg3)
+
+    assert_mock mock
+  end
```
zenspider, line 288: I've adopted this test and expanded it into 9 edge cases (block vs args, missing vs extra, bad keys, bad vals) covering what I think is the whole gamut. Thank you for the nudge!

lib/minitest/mock.rb
```diff
@@ -169,6 +169,8 @@ def method_missing sym, *args, &block # :nodoc:
       retval
     end
 
+    ruby2_keywords :method_missing if respond_to?(:ruby2_keywords, true)
```
zenspider, line 172: Unfortunately while this is a nice small diff, it doesn't really cover what needs to be done for kwargs to be fully supported. I think I've got that in place now, but I'd love your eyes on my changes to see if I missed anything.

zenspider:
As I said in the review. the tests are a good start and I've borrowed ideas from it. I think I have the implementation covered so I'm going to close this PR. Thank you for the contribution and the nudge!

benortiz:
Thanks for expanding the implementation and coverage and taking this over the finish line! I appreciate it very much 🙇🏻

PR #908, opened 2022-05-11
fix: make Mock#expect support keyword arguments

### LINKED ISSUE #847: mock: Keyword arguments deprecation warning when calling block
## Example Code
```ruby
m = Minitest::Mock.new

m.expect(:foo, nil) do |example:|
  example == :bar
end

m.foo(example: :bar)

assert m.verify
```

## Expected

When calling passing a block to `Mock#expects`, Minitest should call the block, and not produce any Ruby warnings.

## Actual

Results in a keyword parameters deprecation warning when the mock framework calls the block.
```
./usr/local/bundle/gems/minitest-5.14.0/lib/minitest/mock.rb:144: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

## Versions
Ruby 2.7.1
Minitest 5.14.0

tonytonyjan:
This change aims to fix the bug derived from the following code:

```
m = Minitest::Mock.new
m.expect(:foo, nil) { |bar:| bar == 'bar' }
m.foo(bar: 'bar')
```

Keyword arguments are separated from other arguments since Ruby v3. The current implementation of `Minitest::Mock#method_missing` would print warning message in Ruby v2.7 and no longer work in Ruby v3.

## Bug Reproduce

It shows deprecation warning message In Ruby 2.7.6:

```sh
docker run --rm -i ruby:2.7.6-alpine3.15 sh -c \
  'gem install --silent minitest:5.15.0 && exec "$@"' sh \
  ruby -W:deprecated -rminitest/mock <<EOS
m = Minitest::Mock.new
m.expect(:foo, nil) { |bar:| bar == 'bar' }
m.foo(bar: 'bar')
EOS
```

```
/usr/local/bundle/gems/minitest-5.15.0/lib/minitest/mock.rb:144: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
-e:1: warning: The called method `call' is defined here
```

It raises `ArgumentError` in Ruby 3.1.2:

```sh
docker run --rm -i ruby:3.1.2-alpine3.15 sh -c \
  'gem install --silent minitest:5.15.0 && exec "$@"' sh \
  ruby -W:deprecated -rminitest/mock <<EOS
m = Minitest::Mock.new
m.expect(:foo, nil) { |bar:| bar == 'bar' }
m.foo(bar: 'bar')
EOS
```

```
-e:1:in `block in <main>': missing keyword: :bar (ArgumentError)
        from /usr/local/lib/ruby/gems/3.1.0/gems/minitest-5.15.0/lib/minitest/mock.rb:144:in `method_missing'
        from -e:1:in `<main>'
```

fix #847

### DIFF
--- a/test/minitest/test_minitest_mock.rb
+++ b/test/minitest/test_minitest_mock.rb
@@ -210,7 +210,7 @@ class TestMinitestMock < Minitest::Test
       mock.a
     end
 
-    assert_equal "No more expects available for :a: []", e.message
+    assert_equal "No more expects available for :a: [] {}", e.message
   end
 
   def test_same_method_expects_are_verified_when_all_called
@@ -286,6 +286,13 @@ class TestMinitestMock < Minitest::Test
     assert_mock mock
   end
 
+  def test_mock_forward_keyword_arguments
+    mock = Minitest::Mock.new
+    mock.expect(:foo, nil) { |bar:| bar == 'bar' }
+    mock.foo(bar: 'bar')
+    assert_mock mock
+  end
+
   def test_verify_fails_when_mock_block_returns_false
     mock = Minitest::Mock.new
     mock.expect :foo, nil do
@@ -293,7 +300,7 @@ class TestMinitestMock < Minitest::Test
     end
 
     e = assert_raises(MockExpectationError) { mock.foo }
-    exp = "mocked method :foo failed block w/ []"
+    exp = "mocked method :foo failed block w/ [] {}"
 
     assert_equal exp, e.message
   end
--- a/lib/minitest/mock.rb
+++ b/lib/minitest/mock.rb
@@ -115,10 +115,10 @@ module Minitest # :nodoc:
       true
     end
 
-    def method_missing sym, *args, &block # :nodoc:
+    def method_missing sym, *args, **kwargs, &block # :nodoc:
       unless @expected_calls.key?(sym) then
         if @delegator && @delegator.respond_to?(sym)
-          return @delegator.public_send(sym, *args, &block)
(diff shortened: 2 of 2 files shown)


icheishvili (approved):
`**kwargs` is the more idiomatic style that I've seen over `**keywords`, but otherwise this looks good. I'm quite interested in this because it's stopping us from upgrading several projects to Ruby 3.x.

tonytonyjan:
@icheishvili thank you for the feedback! I've pushed again to follow the idiomatic style by renaming `keywords` to `kwargs`.

icheishvili:
@zenspider Can you grant one of us contributor access? We'd love to help move the project along.

tonytonyjan:
@zenspider お願いします！

test/minitest/test_minitest_mock.rb
```diff
@@ -286,14 +286,21 @@ def test_mock_block_is_passed_function_block
     assert_mock mock
   end
 
+  def test_mock_forward_keyword_arguments
+    mock = Minitest::Mock.new
+    mock.expect(:foo, nil) { |bar:| bar == 'bar' }
+    mock.foo(bar: 'bar')
+    assert_mock mock
+  end
```
zenspider, line 294: this is only testing one small aspect of a rather complex problem.

zenspider:
Unfortunately while this is a nice small diff, it doesn't really cover what needs to be done for kwargs to be fully supported. I think I've got that in place now, but I'd love your eyes on my changes to see if I missed anything. Thank you for your PR!

santucorephp:
good thought

Issue #895, opened 2022-01-25
RuntimeError: instance variable added to NameError instance


artemave:
After this change was merged https://github.com/seattlerb/minitest/commit/ad135e9def3d9dd9264e0cfa82c85f8f86f6ade1 _some_ of our test errors are being masked by the following one:

```
Error:
TasksControllerTest#around_all:
RuntimeError: instance variable added to NameError instance
```

For example, we'd get that when trying to use a non-existent method.

With `Marshal.dump` commented out, the actual test error is shown as expected. E.g.:

```
NameError: undefined local variable or method `bbbbbb' for #<TasksControllerTest:0x00007fe000496618>
```

I couldn't figure out how to reproduce the error, but the fact remains that in _some cases_ `Marshal.dump` raises with something other than `TypeError`.

Our ruby version is 2.7.5p203.

gstokkink:
We have the same thing occurring here, with Minitest 5.15.0 and Ruby 2.7.5p203.

sokkalf:
We're having the same problem.

We use mechanize in a test, and it occasionally gets a Net::ReadTimeout. In our case, an IOError gets passed to `Marshal.dump` and it crashes. This also stops execution of all following tests, so this is a big problem for us. Only solution we found was to downgrade to 5.14.

```
# Running:                                                                                
                                                                                                                                                                                     
Traceback (most recent call last):                                                        
        36: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:73:in `block in autorun'                                                          
        35: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:146:in `run'
        34: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:169:in `__run'                                                                    
        33: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:169:in `map'
        32: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/minitest-5.15.0/lib/minitest.rb:169:in `block in __run'                                                           
        31: from /home/vagrant/my_project/vendor/bundle/ruby/2.7.0/gems/railties-6.1.4.7/lib/rails/test_unit/line_filtering.rb:10:in `run'                               
[...]

PR #909, opened 2022-05-19
Handle exceptions from Marshal.dump

baelter:
Fixes [#895](https://github.com/minitest/minitest/issues/895)


zenspider:
There's no test to cover the edge cases added in this PR...

I'm not comfortable applying this.

I believe I've handled it in 31b95ed

PR #915, opened 2022-06-21
Update minitest.rb for backward-compatible

mayankesh-maurya:
Minitest upgraded to 5.16 in June 2022 which requires >= 2.6, This is a breaking change and broking existing applications that are using rails 5.2.3 and ruby version >2.5 and <2.6.

ERROR: Error installing rails:
minitest requires Ruby version < 4.0, >= 2.6. The current ruby version is 2.5.0.

According to Versioning.
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backward-compatible manner, and
PATCH version when you make backward-compatible bug fixes.

As it makes incompatible API changes, it should be a major release as we are moving from ruby 2.2>= to 2.6>=.

Please approve changes.

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -1,8 +1,9 @@
 === 5.16.0 / 2022-06-14
 
-* 1 major enhancement:
+* 2 major enhancements:
 
   * Added Minitest::TestTask.
+  * Dropping ruby 2.2 - 2.5. 2.6 is DTM soon too.
 
 * 11 minor enhancements:
 
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,7 @@ Hoe.spec "minitest" do
 
   license "MIT"
 
-  require_ruby_version [">= 2.2", "< 4.0"]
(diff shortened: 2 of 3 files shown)


zenspider:
How is the ruby version supported an incompatible API change? It isn't an API.

Also: there's nothing forcing you to upgrade minitest. The easiest workaround is to not do that.

Also: rails 5.2 is compatible with ruby 2.6 and the performance benefits are worth the upgrade. I recommend it.

zenspider:
Rails 5.2 EOL'd recently: https://endoflife.date/rails which drops my requirements of supporting ruby 2.2 - 2.6. I kept 2.6 around simply because my work requires it still (tho that will change in a few weeks).

Ruby 2.5 EOL'd > 1 year ago: https://endoflife.date/ruby
Ruby 2.6 EOL'd recently as well.

Luckily, upgrading ruby is pretty easy and uneventful for the most part.

That said, I'm not going to accept this PR.

PR #927, opened 2022-09-18
Remove check for `.b` method

tenderlove:
String#b was added in ruby/ruby@a6ec2584527f and shipped with Ruby v2.0.0 and up.  I think we can stop checking for it now


zenspider:
THAT IS CRAZY TALK

zenspider:
done. ❤️

PR #933, opened 2022-11-16
Send assertion kwargs to expectations

blowmage:
Another change needed to support keyword arguments on Ruby 3.

This will fix minitest/minitest-rails#246


zenspider:
I've taken your tests, but I think the implementation is too fragile across versions of ruby. 

See 87604fca for implementation and https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html for explanation.

blowmage:
:+1:

PR #934, opened 2022-11-19
Fix race condition causing flaky tests

XrXr:
`TestMinitestMock` used to be parallelized but
it uses `with_kwargs_env()` which changes an environment variable.
Environment variable is global to the process so there are windows
where the variable is deleted by another thread before the test
is done running.

Run the tests serially to fix the occasional failure.

---
We run these tests on CI over at ruby/ruby and sometimes I see failures coming from this issue.


zenspider:
GOOD CATCH!

zenspider:
done. thanks!

PR #936, opened 2022-11-22
Introduce `assert_pattern` for pattern matching

flavorjones:
## Summary

For testing with pattern matching (only supported with Ruby 3.0 and later)

``` ruby
# pass
assert_pattern { [1,2,3] => [Integer, Integer, Integer] }

# fail "length mismatch (given 3, expected 1)"
assert_pattern { [1,2,3] => [Integer] }
```

## Use case

I can imagine a test case like this exists today:

``` ruby
def test_the_results
  objs = call_that_returns_something
  assert_equal(2, objs.length)
  assert_equal("bob", objs.first.name)
  assert_equal(24, objs.first.age)
  assert_equal("jim", objs.last.name)
  assert_equal(79, objs.last.age)
end
```

I think it's far less verbose and more readable to write:

``` ruby
def test_the_results
  assert_pattern do
    call_that_returns_something => [{name: "bob", age: 24}, {name: "jim", age: 79}] }
  end
end
```

### `assert` with `<expression> in <pattern>`

One alternative might be to use `assert` along with `<expression> in <pattern>`, like this:

``` ruby
assert(( [1,2,3] in [Integer, Integer, Integer] ))
```

Two notes:

- Yes, we need to use _two_ parens to do this. Yes, that's ugly. See https://bugs.ruby-lang.org/issues/19033 for background.
- No descriptive error message is available, since `<expression> in <pattern>` is a boolean expression. Using `<expression> => <pattern>` raises an exception with very specific descriptions!

### no assertions, just pattern matching

I can imagine an argument being made that this assertion is pretty close to the idea of `refute_raises` (whose nonexistence I mostly agree with), which would lead us to write:

``` ruby
def test_the_results
  call_that_returns_something => [{name: "bob", age: 24}, {name: "jim", age: 79}]
end
```

The `<expression> => <pattern>` expression by itself will raise a `NoMatchingPatternError` on match failure, which is counted as a test _error_. Using this helper will ensure we can easily distinguish between a runtime error and an assertion failure.

I think it's also helpful to visually distinguish between the invocation of `<expression> => <pattern>` to bind a local variable from an invocation that is serving purely as an assertion.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -358,6 +358,32 @@ module Minitest
     end
 
     ##
+    # For testing with pattern matching (only supported with Ruby 3.0 and later)
+    #
+    #   # pass
+    #   assert_pattern { [1,2,3] => [Integer, Integer, Integer] }
+    #
+    #   # fail "length mismatch (given 3, expected 1)"
+    #   assert_pattern { [1,2,3] => [Integer] }
+    #
+    # The bare <tt>=></tt> pattern will raise a NoMatchingPatternError on failure, which would
+    # normally be counted as a test error. This assertion rescues NoMatchingPatternError and
+    # generates a test failure. Any other exception will be raised as normal and generate a test
+    # error.
+    #
+    def assert_pattern
+      raise NotImplementedError, "only available in Ruby 3.0+" unless RUBY_VERSION >= "3.0"
+      flunk "assert_pattern requires a block to capture errors." unless block_given?
+
+      begin
+        yield
+        pass
+      rescue NoMatchingPatternError => e
+        flunk(e.message)
+      end
+    end
+
+    ##
     # For testing with predicates. Eg:
     #
     #   assert_predicate str, :empty?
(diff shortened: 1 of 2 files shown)


zenspider:
I wish  I'd looked at this a little more closely a couple weeks ago. Now that I do, I see that it isn't QUITE ready and that I shouldn't sneak it in just before midnight.... Soon tho. I like the idea.

zenspider:
It really just needed to use `eval` where you have the "imagine" comments.... I have something on the way

zenspider:
I sometimes use `refute_match` to show something is missing, do a thing, and then `assert_match` to show that it now exists... 

Do you not see a parallel between `assert_pattern` and `assert_match`? Might a `refute_pattern` be useful? 

I'm totally neutral on this one and just asking.

flavorjones:
> It really just needed to use `eval` where you have the "imagine" comments.... I have something on the way

Just force-pushed that change.

Thinking about `refute_pattern`, let me play with it for a few minutes to see if it feels useful.

flavorjones:
OK, I've also added `refute_pattern`.

flavorjones:
@zenspider Looks like CI is failing because it can't find `minitest/test_task` with Hoe 4. Not sure if your intention is to use the test task that's under test, or pull in an older Hoe, or pull in an already-released minitest, so I'm going to leave it to you to fix up (hope that's OK).

zenspider:
```
Passed: 3.1.3, 3.2.0
Failed: 2.6.10, 2.7.7, 3.0.5
```

I'm poking...

zenspider:
FFS, yuck:

```ruby
    exp_msg = RUBY_VERSION >= "3.1" ? /length mismatch/ : "[1, 2, 3]" # WTF?
```

zenspider:
OK! I just pushed this with some fixes and extensions to use it in minitest/spec. Thank you so much! Sorry I dragged my feet on this. I'll get a release out shortly.

flavorjones:
Hey, thank you for the extra effort to clean this up and get it merged! :heart:

PR #938, opened 2022-11-29
Add ability to match exception messages to `assert_raises` assertion

fatkodima:
Instead of this
```ruby
error = assert_raises(SomeError) do
  do_something
end
assert_equal "Some error message", error.message
```

it is now possible to write this
```ruby
assert_raises(SomeError, match: "Some error message") do # or using a regexp
  do_something
end
```

It is a very popular pattern in tests. I found 373 matches of it in the Rails codebase itself.

This change will also help to avoid a common pitfall when people pass the last argument (as a string/regexp) to `assert_raises` thinking that it will be matched against the error too. There is even a rubocop cop for this - https://github.com/rubocop/rubocop-minitest/blob/master/lib/rubocop/cop/minitest/assert_raises_with_regexp_argument.rb. Personally, I was in this trap several times.

I opened a discussion within the rails community - https://discuss.rubyonrails.org/t/add-assert-raises-with-message-testing-helper/81831 and people find it useful.

I also opened a similar PR in the rails itself - https://github.com/rails/rails/pull/46611, in case this won't be merged. People from the rails core agree on merging this, but we think it is better to upstream to minitest.

There was a similar issue/PR some time ago, supported by many within comments - https://github.com/minitest/minitest/issues/530, but with a different implementation.

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -392,8 +392,15 @@ module Minitest
     #   end
     #
     #   assert_equal 'This is really bad', error.message
+    #
+    # Accepts an optional +match+ matcher to test against the raised
+    # exception message. Eg:
+    #
+    #   assert_raises(ArgumentError, match: /bad argument/i) do
+    #     raise ArgumentError, 'Bad argument'
+    #   end
 
-    def assert_raises *exp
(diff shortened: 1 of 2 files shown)


zenspider:
Dupe. No thank you.

PR #941, opened 2023-01-01
Adds Ruby 3.2 to the CI matrix

petergoldstein:
Also updates the checkout action versions.

To get older Rubies (pre-Ruby 3.2) to run green, I had to add an install minitest step to ensure that the latest minitest is used, rather than the one bundled with those Rubies.

This all runs green on my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
           - '2.7'
           - '3.0'
           - '3.1'
+          - '3.2'
           - head
         include:
           - os: windows-latest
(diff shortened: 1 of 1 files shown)


zenspider:
>  I had to add an install minitest step to ensure that the latest minitest is used, rather than the one bundled with those Rubies.

I've addressed this by changing the load path in the Rakefile, so it is the CURRENT version, rather than the last published version.

zenspider:
I pulled in your first commit. Also added 2.6 to reflect reality. Thanks!

PR #945, opened 2023-03-06
Use minitest organization

hsbt:
👋 https://rubygems.org/gems/minitest still used old organization name. I fixed it.


zenspider:
done. thank you!

PR #950, opened 2023-05-05
Extend the class documentation a bit

murb:
I am often browsing documentation in Dash; which makes you jump quickly to documentation on class/module level. Minitest documentation on that level, however, lacks simple instructions as given in the README. I simply copied those to the classes. Hope it helps & is appreciated :) Otherwise, feel free to dismiss it.


zenspider:
This feels too redundant to me. Closing.

PR #951, opened 2023-05-19
Replace `MiniTest` with `Minitest` in example code

sambostock:
This replaces the usage of `MiniTest::Test` in a `LifecycleHooks` documentation comment code example with `Minitest::Test`.

There doesn't appear to be a reason for it to use the old `MiniTest` naming, so it must just have been missed when the switch was made.


zenspider:
Done! Thank you!

PR #959, opened 2023-07-13
Add `default:` kwarg to `TestTask#create` & `#define`

### LINKED ISSUE #958: Make `TestTask.create` adding itself as a `default` task dependency optional
`TestTask.create` builds a variety of Rake tasks and are much appreciated.

However, it is mandartory and un(explicitly )documented that it also modifies the `default` task, which may be undesired if the `default` is for another purpose.

For example, I’m developing a gem that uses a `.gemspec`-registered Rakefile to build (no, it’s not a C extension, which would instead use an `extconf.rb`). I’m not interested to force users to run tests after `gem install`.

ParadoxV5:
Closes #958

Using a keyword arg rather than a positinal arg since its use isn’t frequent. The kwarg is optional and defaults to `true` to retain current behavior. (Forward compatibility)

Don’t worry, kwargs are available in old Rubies too. Although hopefully nobody’s relying on that `TestTask` doesn’t accept kwargs… (*their* `ruby2_keywords` problem)

### DIFF
--- a/lib/minitest/test_task.rb
+++ b/lib/minitest/test_task.rb
@@ -34,14 +34,15 @@ module Minitest # :nodoc:
     WINDOWS = RbConfig::CONFIG["host_os"] =~ /mswin|mingw/ # :nodoc:
 
     ##
-    # Create several test-oriented tasks under +name+. Takes an
-    # optional block to customize variables.
+    # Create several test-oriented tasks under +name+ (default +test+).
+    # Also adds a test task to the default task(s) unless +default: false+.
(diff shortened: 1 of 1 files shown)


ParadoxV5:
While waiting for the CI to check d7d7c0f, I looked into `test_minitest_test_task.rb`, and realized that only `TestTask#make_test_cmd` is tested (which is unrelated to this issue).
Therefore, I was hesitant to write tests for the PR (which requires researching how Rake tests their DSL).

zenspider:
Closing this in favor of 47959978

PR #964, opened 2023-08-30
Use inline code tags in History file to prevent github from auto-linking text near ENV

DimaD:
Github tries to convert text like `ENV["MT_COMPAT"]` into a link which leads nowhere. This PR fixes 2 most recent occurrences of this in the History file.

Before the change:
minitest 2023-08-30 13-57-06

After the change:

minitest 2023-08-30 13-57-41


zenspider:
Dupe of #979. Already fixed. Sorry

PR #968, opened 2023-09-11
Add changelog link to README and gem specification

mattbrictson:
I noticed that minitest lacks a link to its changelog on its [rubygems.org page](https://rubygems.org/gems/minitest).

This commit adds a changelog link to the README using the special `clog` notation. This URL will automatically be picked up by `hoe`[^1] and be  included in the gem specification metadata the next time the gem is published.

 [^1]: https://github.com/seattlerb/hoe/blob/bc130abf66c8447254e2ac2e070970e40c0a0158/lib/hoe.rb#L116


zenspider:
done! Thanks!

PR #973, opened 2023-10-19
Removed thread and mutex_m dependencies

hsbt:
Ruby core team have plan to extract `mutex_m` from stdlib at Ruby 3.4. After that, `minitest` needs to depend `mutex_m` gem.

I replaced `mutex_m` with simple `Mutex` usage. But I'm not familiar with thread programming. It may be wrong usage.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -1,5 +1,4 @@
 require "optparse"
-require "thread"
 require "mutex_m"
 require "minitest/parallel"
 require "stringio"
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -1,5 +1,4 @@
 require "optparse"
-require "mutex_m"
 require "minitest/parallel"
 require "stringio"
 require "etc"
@@ -618,8 +617,6 @@ module Minitest
   # you want. Go nuts.
 
   class AbstractReporter
-    include Mutex_m
-
     ##
     # Starts reporting on the run.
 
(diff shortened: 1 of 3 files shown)


lib/minitest.rb
```diff
@@ -655,6 +656,10 @@ def report
     def passed?
       true
     end
+
+    def mutex # :nodoc:
+      @mutex
+    end
```
eregon, line 662:
```suggestion
    def synchronize(&block) # :nodoc:
      @mutex.synchronize(&block)
    end
```
This would be cleaner interface-wise and avoid exposing the Mutex.
Also it would avoid needing changes in lib/minitest/parallel.rb

hsbt: 👍  Thanks. It's better than my changes.

zenspider:
DONE! Thank you. Sorry for the delay. I'll get this out in the next release shortly

hsbt:
@zenspider Thank you!

Earlopain:
@zenspider hope you don't mind the ping. When do you except to cut the next release? Ruby main has now removed mutex_m code and as such testing against ruby-head errors unless manually adding it to the gemfile/gemspec.

zenspider:
Just released v5.21.0 with this fix.

Earlopain:
Thank you, I appreciate that very much.

hsbt:
Thanks always!

PR #975, opened 2023-10-31
Add missing `include_all` param to `assert_respond_to`

larouxn:
Add `include_all` param to `assert_respond_to` to mirror `Object#respond_to?`'s parameters.

https://docs.ruby-lang.org/en/3.2/DRb/DRbObject.html#method-i-respond_to-3F

~Am aware this will be a breaking change for anyone with a message specified as people will need to specify `true` or `false` for `include_all` in order to use the `msg` parameter as everything is positional. Alternatively could use some keywords arguments but that's not done much (if at all) in Minitest.~

~Open to suggestions on how best to add `respond_to?`'s missing parameter to `assert_respond_to`.~

Uses a keyword argument for the new `include_all` parameter to allow for backwards compatibility (no breakage).

### DIFF
--- a/lib/minitest/assertions.rb
+++ b/lib/minitest/assertions.rb
@@ -449,12 +449,13 @@ module Minitest
 
     ##
     # Fails unless +obj+ responds to +meth+.
+    # include_all defaults to false to match Object#respond_to?
 
-    def assert_respond_to obj, meth, msg = nil
+    def assert_respond_to obj, meth, include_all: false, msg = nil
       msg = message(msg) {
         "Expected #{mu_pp(obj)} (#{obj.class}) to respond to ##{meth}"
       }
-      assert obj.respond_to?(meth), msg
+      assert obj.respond_to?(meth, include_all), msg
     end
 
     ##
--- a/test/minitest/test_minitest_assertions.rb
+++ b/test/minitest/test_minitest_assertions.rb
@@ -41,6 +41,12 @@ class TestMinitestAssertions < Minitest::Test
       self.assertions = 0
       self.failure = nil
     end
+
+    private
+
+    def private_method
+      puts "Hello, I'm private"
+    end
   end
 
   def setup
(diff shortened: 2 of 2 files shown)


lib/minitest/assertions.rb
```diff
@@ -449,12 +449,13 @@ def assert_raises *exp
 
     ##
     # Fails unless +obj+ responds to +meth+.
+    # include_all defaults to false to match Object#respond_to?
 
-    def assert_respond_to obj, meth, msg = nil
+    def assert_respond_to obj, meth, include_all = false, msg = nil
```
zenspider, line 454: make `include_all` a kwarg to prevent the breakage

zenspider (changes requested):

lib/minitest/assertions.rb
```diff
@@ -449,12 +449,13 @@ def assert_raises *exp
 
     ##
     # Fails unless +obj+ responds to +meth+.
+    # include_all defaults to false to match Object#respond_to?
 
-    def assert_respond_to obj, meth, msg = nil
+    def assert_respond_to obj, meth, include_all: false, msg = nil
```
zenspider, line 454: this isn't syntactically correct

AND it fails the tests once fixed

zenspider (changes requested):

zenspider:
This is just broken in many different ways. I'm gonna shut this down and I'll fix it.

larouxn:
Apologies, I have it fixed (https://github.com/minitest/minitest/commit/bdebcbb2965b8d8babfad4cc1fe2735d5c828eae) but I see you already merged support in https://github.com/minitest/minitest/commit/f908dda892e70b0389c61c12137a18be8e3fb179. Much appreciated.
Screenshot from 2023-11-11 09-25-03

lib/minitest/assertions.rb
larouxn: Apologies, I have it fixed (https://github.com/minitest/minitest/commit/bdebcbb2965b8d8babfad4cc1fe2735d5c828eae) but I see you already merged support in https://github.com/minitest/minitest/commit/f908dda892e70b0389c61c12137a18be8e3fb179. Much appreciated.
Screenshot from 2023-11-11 09-25-03

PR #977, opened 2023-11-17
Do not package test files

pusewicz:
Test files are generally not accessible when installing gems. We can cut them so that we can save some bytes when transferring and storing on all the machines this gem gets installed to.

I did search for a possible PR/issue that was mentioning the above, but could not find anything.

If we go this route, I imagine we could also remove the `.autotest`, `Rakefile` and possibly the `MANIFEST.txt` file.

Please let me know what you think!


zenspider:
no

PR #979, opened 2023-12-25
Fix rdoc to prevent wrong links and wrong quotes

stomar:
Escape `ENV[...]` and `option[...]`, which else is interpreted as link; also escape some quote characters to prevent wrong "smart" quotes.


zenspider:
done! thanks!

PR #991, opened 2024-03-15
Make the test suite compatible with --enable-frozen-string-literal

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205

Since `minitest` is tested as part of ruby-core CI, it needs to be compatible with the `--enable-frozen-string-literal` option.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,9 @@ jobs:
             ruby: ruby
           - os: macos-latest
             ruby: ruby
+          - os: ubuntu-latest
+            ruby: ruby
+            rubyopt: "--enable-frozen-string-literal"
 
     steps:
       - name: repo checkout
@@ -76,5 +79,5 @@ jobs:
 
       - name: rake test:isolated
         run: |
-          rake test:isolated
+          rake test:isolated RUBYOPT="${{ matrix.rubyopt }}"
         timeout-minutes: 3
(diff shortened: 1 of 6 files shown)


zenspider:
Coulda sworn this was already done... 

Why do you seem to prefer dup over unary `+` in some places?

byroot:
Fir chaining. I prefer to avoid (+"str").method.

But happy to change if you have sone preference. It's mostly just style.

.github/workflows/ci.yml
```diff
@@ -26,10 +26,14 @@ jobs:
           - '3.0'
           - '3.1'
           - '3.2'
+          - '3.3'
```
zenspider, line 29: I'm gonna steal this line right away and it should disappear after a rebase... leave the inclusion below tho

casperisfine: Rebased.

casperisfine:
Sorry to be a bother, but is there anything blocking the merge I could fix?

This is one of the last two blockers for https://github.com/ruby/ruby/pull/10262

casperisfine:
I've been told you are in vacation, so apologies for the previous ping, and I'll figure a workaround.

zenspider:
nah! I got you

zenspider:
OK. This is done! Do you need a release? I assume not since this is test only but maybe you need a tag?

byroot:
No release nor tag needed. Thank you!

Issue #996, opened 2024-05-02
Minitest references Rails in non-Rails code


soulcutter:
Given a file "minitest.rb" that requires minitest/autorun
And minitest 5.22.3 is being used
When I run ruby -Ilib:test ./minitest.rb

I get the output on a failing test:

bin/rails test ./minitest.rb:5

This file does not require any rails, there is no Gemfile, AFAIK nothing that might load any rails thing. bin/rails does not exist at this location. I would expect there be no reference to Rails.

Here's `minitest.rb`
```ruby
require 'minitest/autorun'

class MyTestThatProducesADiff < Minitest::Test

  def test_diff
    assert_equal "Mayonnaise is delicious", "Mayonnaise is delicioous"
  end
end
```

Output
```
$ ruby -Ilib:test ./minitest.rb
Run options: --seed 43692

# Running:

F

Failure:
MyTestThatProducesADiff#test_diff [./minitest.rb:6]:
Expected: "Mayonnaise is delicious"
  Actual: "Mayonnaise is delicioous"


bin/rails test ./minitest.rb:5



Finished in 0.000435s, 2298.8505 runs/s, 2298.8505 assertions/s.
1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

In case it might be getting polluted from something automatically trying to require rails, here's a filtered list of what gems are present (but not required in my code):

```
$ gem list | grep rails
autoprefixer-rails (10.4.16.0)
dotenv-rails (3.1.0)
factory_bot_rails (6.4.3)
pry-rails (0.3.9)
rails (7.1.3.2)
rails-dom-testing (2.2.0)
rails-html-sanitizer (1.6.0)
rspec-rails (6.1.2)
ruby-lsp-rails (0.3.5, 0.3.3)
sass-rails (6.0.0)
sassc-rails (2.1.2)
sprockets-rails (3.4.2)

**Comment by blowmage** [comment]
Minitest plugin system looks at all installed gems, so its possible you have a plugin installed that is altering the behavior.

Here is what your test looks like in a clean gemset on my system:

```
 >> chruby 3.3.1

 >> ruby -v
ruby 3.3.1 (2024-04-23 revision c56cd86388) [arm64-darwin23]

 >> omg clean
Switching gemset to clean

 >> gem install minitest
Fetching minitest-5.22.3.gem
Successfully installed minitest-5.22.3
1 gem installed

[...]

PR #997, opened 2024-05-03
Idea for manually registering plugins

tenderlove:
This is just a straw man commit.  It introduces a method called `Minitest.add_plugin`.  You pass an object to `add_plugin` and that object is registered as a plugin.

This commit converts Minitest::PrideIO to use this interface, but that change is only for demonstration purposes.  The existing plugin infrastructure should be backwards compatible.

The advantages of this change are:

1. Plugins don't need to monkeypatch Minitest to register themselves
2. Plugins can manually be added (IOW we don't have to scan gems)

Here's a sample dummy test + plugin:

```ruby
require "minitest/autorun"

# This class would be distributed in a gem
module AaronPlugin
  def self.register!
    Minitest.add_plugin self
  end

  def self.plugin_mt_options opts, options
    # process options
    puts "processing some options"
  end

  def self.plugin_mt_init options
    # do init stuff
    puts "init call"
  end
end

# This could be called via something like:
#
# require "aaron_plugin/autorun"
#
# Or something like this:
#
# require "aaron_plugin"
# AaronPlugin.register!
#
# I'm just manually registering it here
AaronPlugin.register!

# Requiring pride plugin works as usual
require "minitest/pride_plugin"

class FooTest < Minitest::Test
  def test_foo
    assert true
  end
end
```

Running the above test looks like this:

Screenshot 2024-05-02 at 5 08 55 PM

Note that PrideIO works the same, and both runs output the puts calls from my manually registered plugin.

Ref: https://github.com/minitest/minitest/issues/996

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -99,9 +99,9 @@ module Minitest
   end
 
   def self.init_plugins options # :nodoc:
-    self.extensions.each do |name|
+    self.extensions.each do |recv, name|
       msg = "plugin_#{name}_init"
-      send msg, options if self.respond_to? msg
+      recv.send msg, options if recv.respond_to? msg
     end
   end
 
@@ -119,8 +119,12 @@ module Minitest
       seen[name] = true
 
       require plugin_path
-      self.extensions << name
+      self.extensions << [self, name]
     end
+  end
+
+  def self.add_plugin plugin
+    self.extensions << [plugin, "mt"]
   end
 
   ##
@@ -257,9 +261,9 @@ module Minitest
         opts.separator ""
         opts.separator "Known extensions: #{extensions.join(", ")}"
 
-        extensions.each do |meth|
+        extensions.each do |recv, meth|
           msg = "plugin_#{meth}_options"
-          send msg, opts, options if self.respond_to?(msg)
+          recv.send msg, opts, options if recv.respond_to?(msg)
         end
       end
 
--- a/lib/minitest/pride_plugin.rb
+++ b/lib/minitest/pride_plugin.rb
@@ -1,27 +1,29 @@
 require "minitest"
 
 module Minitest
-  def self.plugin_pride_options opts, _options # :nodoc:
-    opts.on "-p", "--pride", "Pride. Show your testing pride!" do
-      PrideIO.pride!
-    end
-  end
-
-  def self.plugin_pride_init options # :nodoc:
-    if PrideIO.pride? then
-      klass = ENV["TERM"] =~ /^xterm|-256color$/ ? PrideLOL : PrideIO
-      io    = klass.new options[:io]
-
-      self.reporter.reporters.grep(Minitest::Reporter).each do |rep|
-        rep.io = io if rep.io.tty?
-      end
-    end
-  end
-
   ##
   # Show your testing pride!
 
   class PrideIO
+    Minitest.add_plugin self
+
+    def self.plugin_mt_options opts, _options # :nodoc:
+      opts.on "-p", "--pride", "Pride. Show your testing pride!" do
+        PrideIO.pride!
+      end
+    end
+
+    def self.plugin_mt_init options # :nodoc:
+      if PrideIO.pride? then
+        klass = ENV["TERM"] =~ /^xterm|-256color$/ ? PrideLOL : PrideIO
+        io    = klass.new options[:io]
+
+        Minitest.reporter.reporters.grep(Minitest::Reporter).each do |rep|
+          rep.io = io if rep.io.tty?
+        end
+      end
+    end
+
     ##
     # Activate the pride plugin. Called from both -p option and minitest/pride
 


adam12:
I like the idea of this. I just had a scenario a few days ago (which @zenspider has already seen) where `railties` being installed globally was causing `ruby -rminitest/autorun -e true` to fail in an entirely unrelated project.

I know there are some other places where the autoloading of plugins is [unfavourable](https://github.com/jeremyevans/sequel/blob/154923695f4dbc4fb2813e5b0a2cc8028113bb96/spec/guards_helper.rb#L1) for the project author.

tenderlove:
I sent [a pr](https://github.com/rails/rails/pull/51718) so that railties will hopefully stop changing minitests output. With this change I think we could convert the railties plugin to be a manual one and that would prevent other environments from loading it in the first place.

zenspider:
Initially I looked at the diff and spit out my coffee... but [skipping whitespace diffs](https://github.com/minitest/minitest/pull/997/files?diff=unified&w=1) makes it much more readable.

zenspider:
@tenderlove your ability to code is never in question... but your ability (or willingness) to consider compatibility absolutely 🌬️ 🐐. 😘

tenderlove:
> @tenderlove your ability to code is never in question... but your ability (or willingness) to consider compatibility absolutely 🌬️ 🐐. 😘

I don't understand 😅

This should be backwards compatible.  It does change `extensions` to store a list of tuples, but if we're worried about external code reading that list, we could handle that too.

blowmage:
I would love to see the plugin system change to use explicit requires like proposed here, and [dropping the gem scanning](https://github.com/minitest/minitest/blob/v5.22.3/lib/minitest.rb#L115-L123) entirely. Is this a change that would be considered for a 6.0 release?

zenspider:
What I'm finding surprising:

```
9974 $ hyperfine ./manual.rb ./automatic.rb
Benchmark 1: ./manual.rb
  Time (mean ± σ):      43.1 ms ±   0.4 ms    [User: 35.4 ms, System: 6.3 ms]
  Range (min … max):    42.2 ms …  45.1 ms    65 runs
 
Benchmark 2: ./automatic.rb
  Time (mean ± σ):      45.3 ms ±   0.6 ms    [User: 36.6 ms, System: 7.3 ms]
  Range (min … max):    44.4 ms …  47.9 ms    63 runs
 
Summary
  ./manual.rb ran
    1.05 ± 0.02 times faster than ./automatic.rb
```

zenspider:
Addressed in b2a4b43b

tenderlove:
@zenspider this isn't really what I had in mind, at all.

There are two problems with your patch.

1.  If Rails requires that file then automatic plugin loading doesn't work _at all_. Rails can require that file, then manually tell Minitest to require the Rails plugin, but now no other plugins will be automatically loaded.  Meaning that people who work on Rails apps but also use automatic plugin loading in their app will face backwards incompatibility (ironically 😛).

2. The situation that I want to avoid is **non Rails apps loading the Rails plugin** and this doesn't fix the situation whatsoever.  We _never_ want Minitest to load the Rails plugin outside of a Rails application, and since `Minitest.load` requires that the plugin file have the same file name format as automatic plugin loading, it means that apps using automatic plugin loading (the normal case) will still try to load the Rails plugin (which is what we _do not_ want).

I really prefer my solution because it allows people do have automatic plugin loading _and_ manual plugin loading. It also means we don't have to put the plugin in a special file thus avoiding the auto load problem.

Can you please reopen this PR as your patch doesn't fix the second problem, and also introduces the first problem?

medwards1771:
@zenspider @tenderlove  I'd like to voice my support allowing Minitest to manually register plugins. 

Here's why:
While working in a non-Rails repo with `minitest` as the only gem, running `ruby <test-filename>` spat out a `Gem:ConflictError`:

```
Unable to activate railties-7.0.8.1, because activesupport-7.1.3.3 conflicts with activesupport (= 7.0.8.1) (Gem::ConflictError)
```

The error traces back to `.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/minitest-5.23.1/lib/minitest.rb` in #self.load_plugins, specifically this block of code:
```
Gem.find_files("minitest/*_plugin.rb").each do |plugin_path|
    name = File.basename plugin_path, "_plugin.rb"

    next if seen[name]
    seen[name] = true

    require plugin_path
    self.extensions << name
end
```

As you know, `Gem.find_files("minitest/*_plugin.rb")` finds plugins for ALL gems with a `minitest` sub-dir, then `require`s them. For example, in my local setup, the plugins found are:

```
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/minitest-5.23.1/lib/minitest/pride_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/minitest-5.22.3/lib/minitest/pride_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/minitest-5.20.0/lib/minitest/pride_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/railties-7.0.8.1/lib/minitest/rails_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/simplecov-0.21.2/lib/minitest/simplecov_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/test-prof-1.3.2/lib/minitest/test_prof_plugin.rb
.local/share/mise/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/test-prof-1.3.1/lib/minitest/test_prof_plugin.rb
```

I don't understand why minitest need to require plugins from railties, simplecov, and test_prof.

Another solution could be to change the glob in `Gem.find_files("minitest/*_plugin.rb")` to select only files within `gems/minitest*/`.

zenspider:
See https://github.com/minitest/minitest/blob/master/lib/minitest/manual_plugins.rb

PR #1003, opened 2024-06-23
Fix the error message when an extension is invalid value

y-yagi:
The current message doesn't show what is a problem. This PR tries to improve it.


zenspider:
oh mah gawd... I let that slip through??? hahaha!

PR #1005, opened 2024-07-24
Accept colon style Hash#inspect in test

tompng:
`Hash#inspect` is proposed to change to `{key: value, non_symbol_key => value}` in https://bugs.ruby-lang.org/issues/20433#note-10
Pull request that change Hash#inspect is here https://github.com/ruby/ruby/pull/10924

This pull request makes test to accept `{key=>value}` style and the proposed `{key => value}` (space around `=>`) style and `{key: value}` style of `Hash#inspect`.

### DIFF
--- a/test/minitest/test_minitest_assertions.rb
+++ b/test/minitest/test_minitest_assertions.rb
@@ -141,10 +141,12 @@ class TestMinitestAssertions < Minitest::Test
     h1, h2 = {}, {}
     h1[1] = Object.new
     h2[1] = Object.new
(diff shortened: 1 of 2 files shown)


zenspider:
You mind if I make changes to make it fit "seattle style" ?

tompng:
Of course I don't mind :+1:

zenspider:
OK. I've modified this a bit ... mostly using `format` + `%p` instead of interpolation in places. Thanks! This'll go out in a bit

PR #1011, opened 2024-08-16
Call a method to run all runnables

tenderlove:
The `run_all_methods` method gets a class, a list of methods to execute, and a reporter.  Then it loops through all methods, running them on each individual test

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -397,9 +397,8 @@ module Minitest
     reset
 
     ##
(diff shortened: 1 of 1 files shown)


jeremyevans:
I think `run_all_methods` should only be called if there are methods to run (e.g. check `empty?` before calling the method instead of inside the method).  Otherwise, this looks good.  I think in either case I could switch to using it.

PR #1017, opened 2024-10-11
Include class name in spec name

thomasmarshall:
This commit changes how spec names are generated to include the test class name. This makes it possible to target all the specs in a class using the `--name` flag.

For example:

```rb
class Foo < Minitest::Spec
  it "a"
  describe "b" do
    it "c"
  end
end
```

| Before | After |
| - | - |
| `Foo#test_0001_a` | `Foo#test_0001_a` |
| `b#test_0001_c` | `Foo::b#test_0001_c` |

Previously only the first spec would be matched by `--name /Foo/` but now they both are. This is also useful if there are multiple classes in the same file. We can select all of the `Bar` tests and none of the `Foo` tests by passing `--name /Bar/`:

```rb
class Foo < Minitest::Spec
  describe "a" do
    it "b"
    it "c"
  end
end

class Bar < Minitest::Spec
  describe "a" do
    it "b"
    it "c"
  end
end
```

Specs outside of classes are still named as before. These will still be named `a#test_0001_b` and `a#test_0001_c`:

```rb
describe "a" do
  it "b"
  it "c"
end
```

I implemented it by pulling out an `is_spec_class` variable and only including the class name if it's a spec class and if the describe stack is empty. It could maybe be a bit prettier but this was a small change.

### DIFF
--- a/lib/minitest/spec.rb
+++ b/lib/minitest/spec.rb
@@ -81,8 +81,9 @@ module Kernel
 
   def describe desc, *additional_desc, &block # :doc:
     stack = Minitest::Spec.describe_stack
-    name  = [stack.last, desc, *additional_desc].compact.join("::")
-    sclas = stack.last || if Class === self && kind_of?(Minitest::Spec::DSL) then
+    is_spec_class = Class === self && kind_of?(Minitest::Spec::DSL)
+    name  = [(self if stack.empty? && is_spec_class), stack.last, desc, *additional_desc].compact.join("::")
+    sclas = stack.last || if is_spec_class then
                             self
                           else
                             Minitest::Spec.spec_type desc, *additional_desc
--- a/test/minitest/test_minitest_spec.rb
+++ b/test/minitest/test_minitest_spec.rb
@@ -952,6 +952,23 @@ class TestMeta < MetaMetaMetaTestCase
     assert_equal "ExampleB::random_method", spec_b.name
   end
 
+  def test_name_inside_class
+    spec_a = nil
+    spec_b = nil
+    inside_class_example = Class.new Minitest::Spec
+    Object.const_set :InsideClassExample, inside_class_example
+    inside_class_example.class_eval do
+      spec_a = describe "a" do
+        spec_b = describe "b" do; end
+      end
+    end
+
+    assert_equal "InsideClassExample::a", spec_a.name
+    assert_equal "InsideClassExample::a::b", spec_b.name
+  ensure
+    Object.send :remove_const, :InsideClassExample
+  end
+
   def test_structure
     x, y, z, * = util_structure
 


zenspider:
If I may ask... why combine your classes and describes in that way at all? What got you to go that route?

thomasmarshall:
Do you consider it misuse to wrap specs manually in a class instead of relying on `register_spec_type` and a top-level `describe`? One example codebase is [Shopify/tapioca](https://github.com/Shopify/tapioca/tree/main/spec) which I believe does this to make Sorbet type checking easier. There may be other reasons, but perhaps it's not that common.

The reason I wanted to target all the tests within a spec class is that the Ruby LSP provides tooling for running tests from the editor. This won't currently select all tests because some will have names like `FooTest#test_0001_a` and others will have names like `b::test_0001_c`.

Screenshot 2024-10-12 at 10 56 40

For regular tests it works great because they all have a common prefix: the class name. I thought by ensuring specs had a common prefix too it would make the above possible.

zenspider:
Huh... It is certainly non-idiomatic and I'm struggling to see the upside. Like you said, "it works great because they all have a common prefix", but the same is true if you do:

```ruby
describe Tapioca do
  ...
```

your original code sample is a counter example where there's an `it` outside the describe, but that top class could be a describe and then it's a done deal... so again, struggling to see the upside.

I think the only place I've seen ruby legit bitch is when someone wants to define a constant inside of a describe block... doesn't matter that I'm sending the whole block over to `Class.new`, it still warns against it... but other than that? dunno!

TO BE CLEAR: I'm not down on your PR... just trying to understand

thomasmarshall:
Yeah that makes sense. I think one reason Tapioca (and other projects using Sorbet) wrap the tests in a class is that Sorbet is doing static type checking so it can't know which class is being dynamically generated by a top-level `describe`. Maybe there are other reasons too but I must admit I don't know.

The reason I proposed this change is not because I necessarily want to write tests this way, but because I previously made some improvements to the built-in minitest integration in Ruby LSP and this part (running all the tests for a spec class) currently doesn't/can't work.

I assumed it was a legit way of using minitest/spec because there are projects in the wild that organize their tests this way, but if it's a misuse or non-idiomatic use of the library, then perhaps it doesn't make sense for Ruby LSP to support that.

zenspider:
I _have_ to assume that sorbet and friends have means to declare types for blocks but that's way outside my purview / care. 

> because there are projects in the wild that organize their tests this way

this is the main concern... there probably are (obviously)... and just because you shouldn't doesn't mean it won't happen. So I like this PR... Imma poke a bit around your refactoring. I never liked that code to begin with and am guessing it could be cleaner before this change goes in.

zenspider:
any changes I make will cause conflict, but I like this change before your idea gets applied:

```diff
@@ -81,14 +81,14 @@
 
   def describe desc, *additional_desc, &block # :doc:
     stack = Minitest::Spec.describe_stack
-    name  = [stack.last, desc, *additional_desc].compact.join("::")
+    name  = [stack.last, desc, *additional_desc]
     sclas = stack.last || if Class === self && kind_of?(Minitest::Spec::DSL) then
                             self
                           else
                             Minitest::Spec.spec_type desc, *additional_desc
                           end
 
-    cls = sclas.create name, desc
+    cls = sclas.create name.compact.join("::"), desc
 
     stack.push cls
     cls.class_eval(&block)
```

then your change is a simple local extraction and a conditional prepend. Something like:

```ruby
    is_spec_class = Class === self && kind_of?(Minitest::Spec::DSL)
    name.prepend self if stack.empty? && is_spec_class
```

still pokin'...

thomasmarshall:
Nice, that looks good. I'll wait until you've finished poking and then I can rebase and deal with the conflict, or feel free to just change my patch directly, whichever you prefer.

zenspider:
after FAR too much mental masturbation, I've settled on the following:
```ruby
    is_spec_class = Class === self && kind_of?(Minitest::Spec::DSL)
    name  = []
    name.append self if stack.empty? && is_spec_class
    name.append stack.last, desc, *additional_desc

    sclas =
      stack.last                 \
      || (is_spec_class && self) \
      || Minitest::Spec.spec_type(desc, *additional_desc)

    cls = sclas.create name.compact.join("::"), desc
```
wdyt?

thomasmarshall:
Nice, I think it looks good. Should I update the PR to include that change?

zenspider:
OK. I've pushed half of the above as a separate commit. This brings your diff down to a beautiful 1 liner and the test.

zenspider:
done! thank you so much for dealing with my questions and delays

thomasmarshall:
Awesome, thanks!

PR #1023, opened 2025-01-03
Reference pattern assertions in expectation comment docs

jaredcwhite:
I noticed the assertion methods in the comment docs don't exist due to typos, so here are the fixed method names.


zenspider:
done! thanks!

I've published the corrected docs. I'll get a gem release out as soon as reasonable.

PR #1024, opened 2025-01-25
Modify ancestors order to allow reaching `Minitest::Assertions#skipped`

Edouard-chin:
Hello 👋 , thanks for maintaining this gem ❤️ . I'd like to propose a change as calling `skipped?` in a test would result in a different output based on where it gets called. 

### Problem

  I'd like to be able to check if a test was skipped in a `ensure` block.

  ```ruby
  def test_something
    skip
  ensure
    puts skipped? #=> nil
  end
  ```

  This is currently not possible as `Minitest::Assertions#skipped?` is not reachable, it will instead reach `Minitest::Reportable#skipped?` which is implemented differently and always return `nil` in a `ensure` block since the test is not yet done and Minitest hasn't set any failures.

### Solution

  I figured the easiest way is to modify the ancestors order on `Minitest::Test` for `Assertions#skipped?` to have precedence.

### DIFF
--- a/lib/minitest/test.rb
+++ b/lib/minitest/test.rb
@@ -9,8 +9,8 @@ module Minitest
 
   class Test < Runnable
     require "minitest/assertions"
-    include Minitest::Assertions
     include Minitest::Reportable
+    include Minitest::Assertions
 
     def class_name # :nodoc:
       self.class.name # for Minitest::Reportable
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -157,6 +157,23 @@ class TestMinitestUnit < MetaMetaMetaTestCase
     refute_predicate test, :skipped?
   end
 
+  def test_skipped_is_reachable
+    test_class = Class.new FakeNamedTest do
+      def test_omg
+        skip
+      ensure
+        flunk unless skipped?
+      end
+    end
+
+    test = test_class.new :test_omg
+    test.run
+
+    refute_predicate test, :error?
+    refute_predicate test, :passed?
+    assert_predicate test, :skipped?
+  end
+
   def util_expand_bt bt
     bt.map { |f| f.start_with?(".") ? File.expand_path(f) : f }
   end


zenspider:
Hi hi... I've reviewed and approve... but I'd love to know, WHY you want such a thing? Why are you checking `skipped?` in an ensure block? (as opposed to a teardown method/after block, for example)

zenspider:
for the record, I tried to have a smaller scope fix by adding `super ||` to the reportable skipped method but that didn't work because it is also used with Result instances which don't have Assertions mixed in. Hits the NotImplementedError in the abstract Runnable#skipped?

Edouard-chin:
Thanks for you reply and accepting the change !

> WHY you want such a thing? 

When there is a test in a suite that requires some particular setup, it may not make sense to add or modify an existing `setup` hook that would apply to the entire suite. So I resort to setting up everyting in the test itself using some local vars (e.g. `old, ENV["VERBOSE"] = ENV["VERBOSE"], "false"`).

I need a way to teardown such tests, but similarly can't use a `teardown` hook as it would apply to the entire suite. I suppose I could extract this test to its own suite, but I'd rather not have to.

```ruby
class SomeTest < Minitest::Test
  # ... Bunch of tests that requires no setup/teardown.

  def test_specific_to_the_java_platform
    skip unless Gem.java_platform?

    old_value = Foo.const_get(:A)
    Foo.cost_set(:A, 42)
  ensure
    return if skipped? 
    Foo.const_set(:a, old_value)
  end
end
```

zenspider:
Hah! I usually make a helper/wrapper method that does your const_set / get for me and I restore it regardless. Something like

```ruby
  def with_const_a val
    old_val = Foo.const_get :A
    Foo.const_set :A, val

    yield
  ensure
    Foo.const_set :A, old_val
  end

  # ...

  def test_blah
    with_const_a 42 do
      # ...
    end
  end
```

zenspider:
done! this will sync up shortly and I'll release it after a bit more triage. Thanks!

PR #1029, opened 2025-02-28
Support Ruby 3.2's Exception#detailed_message feature

neilvcarvalho:
Ruby 3.2 introduced `Exception#detailed_message` to allow gems like `did_you_mean` and `error_highlight` to enrich exception messages. Previously, these gems modified `Exception#message`, which led to confusion in test suites that expected this message to follow a certain pattern. As a result of that change, Minitest stopped showing the enriched message, shown by default when running Ruby programs.

This commit updates Minitest to use `Exception#detailed_message` when available, falling back to message when not, ensuring enriched error messages are displayed without affecting test assertions.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -1084,7 +1084,14 @@ module Minitest
     def message # :nodoc:
       bt = Minitest.filter_backtrace(self.backtrace).join("\n    ")
         .gsub(BASE_RE, "")
-      "#{self.error.class}: #{self.error.message}\n    #{bt}"
+      error_message = if self.error.respond_to?(:detailed_message)
+                        # detailed_message includes he error class name in parentheses
+                        # which we don't want since we're already including the class name
+                        self.error.detailed_message.sub(/ \(#{self.error.class}\)$/, '')
+                      else
+                        self.error.message
+                      end
+      "#{self.error.class}: #{error_message}\n    #{bt}"
     end
 
     def result_label # :nodoc:
--- a/test/minitest/test_minitest_test.rb
+++ b/test/minitest/test_minitest_test.rb
@@ -1371,4 +1371,27 @@ class TestUnexpectedError < Minitest::Test
           c
         EXP
   end
+
+  def test_detailed_message
+    skip "Ruby #{RUBY_VERSION} doesn't support Exception#detailed_message" unless 
+      Exception.instance_methods.include?(:detailed_message)
+
+    exception_class = Class.new(StandardError) do
+      def detailed_message(**options)
+        "Enhanced error message (#{self.class})"
+      end
+
+      def message
+        "Original message"
+      end
+    end
+
+    exception = exception_class.new
+    unexpected_error = Minitest::UnexpectedError.new(exception)
+
+    message = unexpected_error.message
+    assert_includes message, "Enhanced error message"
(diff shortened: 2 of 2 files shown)


skipkayhil:
Hey @zenspider, if you have some time could you take a look at this PR?

This week I spent a bunch of time fixing flaky tests in an application, many of which were caused by `undefined method '[]' for nil:NilClass` type errors. Having `error_highlight` output would have saved me a lot of time tracking down the source of these errors.

lib/minitest.rb
```diff
@@ -1084,7 +1084,14 @@ def backtrace # :nodoc:
     def message # :nodoc:
       bt = Minitest.filter_backtrace(self.backtrace).join("\n    ")
         .gsub(BASE_RE, "")
-      "#{self.error.class}: #{self.error.message}\n    #{bt}"
+      error_message = if self.error.respond_to?(:detailed_message)
+                        # detailed_message includes he error class name in parentheses
+                        # which we don't want since we're already including the class name
+                        self.error.detailed_message.sub(/ \(#{self.error.class}\)$/, '')
```
skipkayhil, line 1090:
```
Error:
TestTest#test_test:
NoMethodError: undefined method '[]' for nil (NoMethodError)

    assert_equal 1, { a: { b: { c: 1 } } }[:a][:ba][:c]
                                                   ^^^^
    test/models/test_test.rb:7:in 'block in <class:TestTest>'
```

I'm not sure the exception class substitution is working. Trying the patch locally I still see the exception class twice.

skipkayhil: Ah, it's because `error_highlight` adds shell escapes to bold part of the error.

```suggestion
                        self.error.detailed_message(highlight: false).sub(/ \(#{self.error.class}\)$/, '')
```

removes the shell escapes but keeps the detailed error. It may be worth raising/rescuing a real error in the test so that `error_highlight`'s behavior can be tested instead of a mock.

neilvcarvalho: Ah, that didn't happen to me, and later I could see that false is the default - perhaps it's configurable, or different exceptions could override that. I'll add that, thanks!

skipkayhil: Oh interesting, I missed the `highlight: false` too. It looks like its defaulting to `true` for me because of `did_you_mean`'s `detailed_message` being called first:

```
irb(Unexpected exception):001> self.error.method(:detailed_message)
=> #<Method: NoMethodError(DidYouMean::Correctable)#detailed_message(highlight: ..., did_you_mean: ...) /home/hartley/.local/share/mise/installs/ruby/3.4.4/lib/ruby/3.4.0/did_you_mean/core_ext/name_error.rb:10>
```

https://github.com/ruby/did_you_mean/blob/50e61993d06b60eecffaf6d36ef34a1de20ad9ba/lib/did_you_mean/core_ext/name_error.rb#L10

neilvcarvalho: Thank you for pointing the issue out and for the suggestion. I changed the implementation a bit with that in mind, and it's working well.

Other tests were failing because of the duplicate class name, but they were just regular exceptions without `error_highlight` or `did_you_mean` involved. I changed the test to trigger `did_you_mean`, which enables `highlight`, and I could finally see a failing test related to ANSI escape sequences.

PR #1039, opened 2025-12-04
Fix reporter test shape variation warning

havenwood:
Running `ruby -W:performance -Ilib:test test/minitest/test_minitest_reporter.rb` on most often hits a random test order that produces a performance warning in Ruby 4. This nit fix just defines instance variables up front so the order is static, avoiding intermittent performance warnings when tests are run.

```
ruby -W:performance -Ilib:test test/minitest/test_minitest_reporter.rb
Run options: --seed 57373

# Running:

...............test/minitest/test_minitest_reporter.rb:68: warning: The class TestMinitestReporter reached 8 shape variations, instance variables accesses will be slower and memory usage increased.
It is recommended to define instance variables in a consistent order, for instance by eagerly defining them all in the #initialize method.
.......
```

### DIFF
--- a/test/minitest/test_minitest_reporter.rb
+++ b/test/minitest/test_minitest_reporter.rb
@@ -29,10 +29,11 @@ class TestMinitestReporter < MetaMetaMetaTestCase
     super
     self.io = StringIO.new(+"")
     self.r  = new_composite_reporter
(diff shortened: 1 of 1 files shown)


zenspider:
I was hoping you found an impl-side performance problem!

thanks! done! should be released soonish™.

PR #1052, opened 2026-01-01
Update README.rdoc replacing rematch with minitest-holdify

ddnexus:
I deprecated rematch, and replaced it with [minitest-holdify](https://github.com/ddnexus/holdify), which is definitely a better way to do the same and more.

This commit updates the changes.


zenspider:
done! thanks! pushed rdoc early to get this out there

PR #1073, opened 2026-05-11
Document that test methods must be public

koic:
As confirmed in https://github.com/rubocop/rubocop-minitest/pull/229, there are real-world cases where non-public methods whose names start with `test_` are silently skipped. This PR clarifies the documentation to make it explicit to users that test methods must be public.

`Minitest::Runnable.methods_matching` uses `public_instance_methods`, so private and protected test_ methods are silently skipped. The behavior was previously implemented but neither documented nor covered by tests, which meant users could accidentally hide tests by changing the visibility of a test method without any signal from the framework.

Clarify the RDoc for both methods_matching and
`Minitest::Test.runnable_methods`. Additionally, since there was no test case for this behavior, add a regression test that locks in the documented behavior so that the visibility filter cannot be accidentally relaxed in the future.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -420,7 +420,8 @@ module Minitest
     end
 
     ##
-    # Returns all instance methods matching the pattern +re+.
+    # Returns all public instance methods matching the pattern +re+.
+    # Methods with private or protected visibility are excluded.
 
     def self.methods_matching re
       public_instance_methods(true).grep(re).map(&:to_s)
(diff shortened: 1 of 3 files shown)


zenspider:
Thank you for bringing this up! I really appreciate it. Sorry this took so long to get to. It required formalizing and publishing some policy and I dragged my feet on that.

Unfortunately, this doesn't pass [the sniff test](https://zenspider.com/ruby/2026/07/my-ai-policy.html).

I've addressed the documentation concerns (and credited you) and will have it published shortly.

PR #1074, opened 2026-05-14
Drop ruby2_keywords delegation workaround

teoljungberg:
As minitest now requires Ruby >= 3.2, the ruby2_keywords shim added for
Ruby 2.6 compatibility is no longer needed. We can now forward keyword
arguments explicitly with `**kws` instead, preserving the existing block
handling where only the `:block` branch passes `&target`.


zenspider:
Finally got to this one! Thanks! Will release "soon".

PR #1075, opened 2026-05-18
Add Minitest.pending? predicate

sferik:
Returns `true` between `Minitest.autorun` being called and `Minitest.run` completing. Lets outside libraries with their own `at_exit` hooks decide whether to route work through `Minitest.after_run` to avoid the `at_exit` LIFO race that drops their results when they were loaded after `minitest/autorun`.

Adds a `@@completed` class variable that `Minitest.run` flips to `true` after the report is emitted, alongside the existing `@@installed_at_exit`.

This is an alternate proposal to https://github.com/minitest/minitest/pull/1076, but you could merge both.

References https://github.com/simplecov-ruby/simplecov/issues/1032.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -13,6 +13,7 @@ module Minitest
   VERSION = "6.0.6" # :nodoc:
 
   @@installed_at_exit ||= false
+  @@completed ||= false
   @@after_run = []
   @extensions = []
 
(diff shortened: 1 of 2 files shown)


zenspider:
While I appreciate the contribution, unfortunately, this doesn't pass [the sniff test](https://github.com/minitest/minitest#ai-policy) and I'm going to close it.

I'll look into enhancing the API around this. Thanks for bringing it up.

zenspider:
In the minitest ecosystem, `Minitest.after_run` is the right way to ensure that your work reports/cleans up/etc after minitest is run. There's no need to complicate it further.

PR #1076, opened 2026-05-18
Add Minitest.autorun_installed? and Minitest.completed? predicates

sferik:
Exposes two pieces of lifecycle state that previously required reaching into the `@@installed_at_exit` class variable:

* `Minitest.autorun_installed?` — public accessor for `@@installed_at_exit` that returns `true` once `Minitest.autorun` has registered its `at_exit` hook.
* `Minitest.completed?` — returns `true` once `Minitest.run` has finished a suite.

Together they answer "is Minitest going to run tests after my at_exit fires?" via `autorun_installed? && !completed?`, which lets outside libraries with their own `at_exit` hooks (e.g. coverage tools loaded after `minitest/autorun`) decide whether to defer their work through `Minitest.after_run` instead of fighting the `at_exit` LIFO ordering.

Adds a `@@completed` class variable that `Minitest.run` flips to `true` after the report is emitted, alongside the existing `@@installed_at_exit`.

This is an alternate proposal to https://github.com/minitest/minitest/pull/1075, but you could merge both.

References https://github.com/simplecov-ruby/simplecov/issues/1032.

### DIFF
--- a/lib/minitest.rb
+++ b/lib/minitest.rb
@@ -13,6 +13,7 @@ module Minitest
   VERSION = "6.0.6" # :nodoc:
 
   @@installed_at_exit ||= false
+  @@completed ||= false
   @@after_run = []
   @extensions = []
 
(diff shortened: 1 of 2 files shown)


zenspider:
While I appreciate the contribution, unfortunately, this doesn't pass [the sniff test](https://github.com/minitest/minitest#ai-policy) and I'm going to close it.

I'll look into enhancing the API around this. Thanks for bringing it up.

zenspider:
I've added `Minitest.installed_at_exit?` to help clarify if autorun has been invoked.

PR #1077, opened 2026-07-01
Make links in readme relative so they work on minite.st too

mamarjan:
These three links don't work on [minite.st docs](https://minite.st/docs/). I'm no expert on rdoc, but I would expect relative links to work, both here and on https://docs.seattlerb.org/minitest/README_rdoc.html


zenspider:
Good catch! Thanks! Changes are already deployed (along with some more)

